PR #409, opened 2022-11-08
Remove `neofs-keywords` dependency and fix `make venv.local-pytest`

aprasolova:
Signed-off-by: anastasia prasolova <anastasia@nspcc.ru>


EliChin (approved):

venv_template.mk
```diff
@@ -4,10 +4,10 @@ venv.$(1): venv.$(1)/bin/activate venv.$(1)/bin/environment.sh
 venv.$(1)/bin/activate: venv/$(1)/requirements.txt submodules
```
abereziny, line 4: Why not remove submodules if we don't have any?

venv/local-pytest/requirements.txt
```diff
@@ -8,3 +8,4 @@ urllib3==1.26.3
 base58==2.1.0
 allure-pytest==2.9.45
 pytest==7.1.2
+pyyaml==6.0
```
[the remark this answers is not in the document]
aprasolova: Ok, I removed `venv/local-pytest/requirements.txt` file.

venv_template.mk
aprasolova: You're right, submodules removed.

PR #413, opened 2022-11-10
Add timeout between commands

anikeev-yadro:
Signed-off-by: anikeev-yadro <a.anikeev@yadro.com>

### DIFF
--- a/robot/resources/lib/python_keywords/node_management.py
+++ b/robot/resources/lib/python_keywords/node_management.py
@@ -188,8 +188,9 @@ def include_node_to_network_map(
 ) -> None:
     node_set_status(hosting, node_to_include, status="online")
 
-    time.sleep(parse_time(MORPH_BLOCK_TIME))
+    time.sleep(parse_time(MORPH_BLOCK_TIME) * 2)
     tick_epoch(shell=shell)
+    time.sleep(parse_time(MORPH_BLOCK_TIME) * 2)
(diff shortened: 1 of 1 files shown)


robot/resources/lib/python_keywords/node_management.py
```diff
@@ -188,8 +188,9 @@ def include_node_to_network_map(
 ) -> None:
     node_set_status(hosting, node_to_include, status="online")
 
-    time.sleep(parse_time(MORPH_BLOCK_TIME))
+    time.sleep(parse_time(MORPH_BLOCK_TIME) * 2)
     tick_epoch(shell=shell)
+    time.sleep(parse_time(MORPH_BLOCK_TIME) * 2)
 
     check_node_in_map(node_to_include, shell, alive_node)
```
abereziny, line 195: It's better and more stable to check some condition in a loop until it becomes true with timeouts instead of fixed time.sleep. 
I have some rough example here:
https://github.com/nspcc-dev/neofs-testcases/pull/411/files#diff-0a0540d91e0e48aca572c160d54e1a13d572219c6e16e43cacf5c538ecaf75fdR18

abereziny (approved):

aprasolova:
@anikeev-yadro I don't see any explanation why you decided to increase the timeout, nor here neither in comments. Roll it back please.

PR #434, opened 2022-11-18
Experimental remove neo3

alexchetaev:


venv/local-pytest/environment.sh
```diff
@@ -1,8 +0,0 @@
-# DevEnv variables
```
abereziny, line 1: why delete this file?

abereziny:
I assume it's outdated. We will continue to use neo3 and new versions also support MacOS.

Issue #312, opened 2022-09-20
Create test for s3 gate. Object


dansingjulia:
- CopyObject
- DeleteObject
- DeleteObjects
- GetObject
- HeadObject
- ListObjects
- ListObjectsV2
- PutObject
- Upload directory

PR #435, opened 2022-11-21
[#312] Add new test for put object from path tree

dansingjulia:
Signed-off-by: Yulia Kovshova <y.kovshova@yadro.com>

### DIFF
--- a/pytest_tests/testsuites/services/s3_gate/test_s3_object.py
+++ b/pytest_tests/testsuites/services/s3_gate/test_s3_object.py
@@ -1,7 +1,8 @@
 import os
+import string
 import uuid
 from datetime import datetime, timedelta
-from random import sample
+from random import choices, sample
 
 import allure
 import pytest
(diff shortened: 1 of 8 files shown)


pytest_tests/testsuites/services/s3_gate/test_s3_object.py
```diff
@@ -915,3 +916,17 @@ def test_s3_sync_dir(self, sync_type):
                 #     "FULL_CONTROL",
                 #     "FULL_CONTROL",
                 # ], "Permission for all groups is FULL_CONTROL"
+
+    @allure.title("Test S3 Put 10 nested level object")
+    def test_s3_put_10_folder(self):
+        path = "/".join(["".join(choices(string.ascii_letters, k=3)) for _ in range(10)])
+        file_path_1 = os.path.join(os.getcwd(), ASSETS_DIR, path, "test_file_1")
```
abereziny, line 923: I was told that we should get rid of ASSETS_DIR var, and we have preprare_tmp_dir fixture which returns 
`os.path.join(os.getcwd(), ASSETS_DIR)`

abereziny (approved):

dansingjulia:
not actual

PR #445, opened 2022-11-30
Add static session token container tests

yadro-vavdeev:
Add static session token container tests

Signed-off-by: Vladimir Avdeev <v.avdeev@yadro.com>


abereziny (approved):
LGTM

Issue #416, opened 2022-11-10
Make fixtures to be entity-named instead of verb-named


abereziny:
Fixtures should reflect entities which they provide instead of verb which they do

PR #450, opened 2022-12-05
Feature abereziny refactor for cluster usage

abereziny:
- Added Cluster object with hierarcy for nodes in a cluster under test.
- Removed hardcoded env variables, now endpoints, passwords and paths are obtained from hosting
- Added base class with all mandatory dependencies
- Fixed some code issues along the way
- #416 

Signed-off-by: Andrey Berezin <a.berezin@yadro.com>

### DIFF
--- a/pytest_tests/helpers/cluster.py
+++ b/pytest_tests/helpers/cluster.py
@@ -1,0 +1,324 @@
+import random
+import re
+from dataclasses import dataclass
+from typing import Any
+
+import data_formatters
+from neofs_testlib.blockchain import RPCClient
+from neofs_testlib.hosting import Host, Hosting
+from neofs_testlib.hosting.config import ServiceConfig
+from test_control import wait_for_success
+
+
+@dataclass
+class NodeBase:
+    """
+    Represents a node of some underlying service
+    """
+
+    id: str
+    name: str
+    host: Host
+
+    def __init__(self, id, name, host) -> None:
+        self.id = id
+        self.name = name
+        self.host = host
+        self.construct()
+
+    def construct(self):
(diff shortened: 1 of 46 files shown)


pytest_tests/helpers/cluster.py
```diff
@@ -0,0 +1,324 @@
...
+    def label(self) -> str:
+        return f"{self.name}: {self.get_endpoint()}"
+
+
+class HTTPGate(NodeBase):
+    """
+    Class represents HTTP gateway in a cluster
+    """
+
+    def get_endpoint(self) -> str:
```
dansingjulia, line 120: we can have a set of endpoins for gate.

pytest_tests/helpers/cluster.py
```diff
@@ -0,0 +1,324 @@
...
+    pass
+
+
+class S3Gate(NodeBase):
+    """
+    Class represents S3 gateway in a cluster
+    """
+
+    def get_endpoint(self) -> str:
+        return self._get_attribute(_ConfigAttributes.ENDPOINT)
```
dansingjulia, line 108: we can have a set of endpoins for gate.

dansingjulia (approved):

pytest_tests/helpers/cluster.py
abereziny: at first we need update testsetup to have multiple addresses in hosting

pytest_tests/helpers/cluster.py
abereziny: at first we need update testsetup to have multiple addresses in hosting

Issue #479, opened 2022-12-22
Could not run tests locally


carpawell:
Got 
```
AssertionError: Regex pattern 'code = 2048.*message = access to object operation denied' does not match "Command 'neofs-cli  storagegroup put --rpc-endpoint 's01.neofs.devenv:8080' --wallet '../neofs-dev-env/services/ir/az.json' --cid '3Y7udEVGbmgu3oq9WcWBQVrkaczMouK9hJysjmyg64vH' --members '48L894MpNdBr4cSCcGTSuQQ8sbj2SkNLumUXmnr47kcL' --lifetime 10' timed out after 30 seconds".
```

Some commands do not have any wallet config (so no password is provided and `neofs-cli` just freezes waiting for a password) while other commands do:

```
15:43:46 [[32mINFO[0m] Command: neofs-cli --config /home/carpawell/NSPCC/git/neofs-testcases/wallet_config.yml storagegroup put --rpc-endpoint 's01.neofs.devenv:8080' --wallet '/home/carpawell/NSPCC/git/neofs-testcases/TemporaryDir/b94d7986-1a5f-4509-b4d9-2c09edb717d4.json' --cid '3Y7udEVGbmgu3oq9WcWBQVrkaczMouK9hJysjmyg64vH' --members '48L894MpNdBr4cSCcGTSuQQ8sbj2SkNLumUXmnr47kcL' --lifetime 10
Success:
return code: 0 
Output: Storage group successfully stored
  ID: 2B1G7qcUm3o2Z5zqWyenbzKmNuuZfHGEkC3t5cayMpdf
  CID: 3Y7udEVGbmgu3oq9WcWBQVrkaczMouK9hJysjmyg64vH

15:43:46 [[32mINFO[0m] Executing command: neofs-cli  storagegroup put --rpc-endpoint 's01.neofs.devenv:8080' --wallet '../neofs-dev-env/services/ir/az.json' --cid '3Y7udEVGbmgu3oq9WcWBQVrkaczMouK9hJysjmyg64vH' --members '48L894MpNdBr4cSCcGTSuQQ8sbj2SkNLumUXmnr47kcL' --lifetime 10
[...]

PR #480, opened 2022-12-22
[#479] Updates for local dev env runs

abereziny:
Covers #479 
Signed-off-by: Andrey Berezin <a.berezin@yadro.com>


carpawell (approved):
Solves the problem mentioned in the issue.

dansingjulia (approved):

PR #484, opened 2022-12-26
Fix: IndexError: list index out of range

vkarak1:
PR fixes the following error:
```
    def choice(self, seq):
        """Choose a random element from a non-empty sequence."""
        # raises IndexError if seq is empty
>       return seq[self._randbelow(len(seq))]
E       IndexError: list index out of range

/usr/lib/python3.9/random.py:347: IndexError
```
Signed-off-by: Vladislav Karakozov <v.karakozov@yadro.com>

### DIFF
--- a/robot/resources/lib/python_keywords/http_gate.py
+++ b/robot/resources/lib/python_keywords/http_gate.py
@@ -286,10 +286,10 @@ def get_object_and_verify_hashes(
         nodes=nodes,
     )
     # for some reason we can face with case when nodes_list is empty due to object resides in all nodes
-    if nodes_list is None:
-        random_node = random.choice(nodes)
-    else:
+    if nodes_list:
         random_node = random.choice(nodes_list)
+    else:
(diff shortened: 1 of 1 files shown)


anikeev-yadro (approved):

robot/resources/lib/python_keywords/http_gate.py
```diff
@@ -286,10 +286,10 @@ def get_object_and_verify_hashes(
         nodes=nodes,
     )
     # for some reason we can face with case when nodes_list is empty due to object resides in all nodes
-    if nodes_list is None:
-        random_node = random.choice(nodes)
-    else:
+    if nodes_list:
         random_node = random.choice(nodes_list)
+    else:
```
abereziny, line 291: This may break the initial intention of this method.
If someone expecting that this method MUST choose nodes without object and we silently ignoring that error may become a hidden issue.

PR #499, opened 2023-01-11
Align nodes epoch

vkarak1:
The code fixes case where expired object is available. The root of the problem was that recently forced epoch was not applied to the node where the object is downloaded.  When `force-new-epoch` command is sent other nodes get this info asynchronously and it needs to wait sometime for alignment.

Signed-off-by: Vladislav Karakozov <v.karakozov@yadro.com>

### DIFF
--- a/pytest_tests/testsuites/services/http_gate/test_http_system_header.py
+++ b/pytest_tests/testsuites/services/http_gate/test_http_system_header.py
@@ -6,15 +6,16 @@ from typing import Optional
 import allure
 import pytest
 from container import create_container
-from epoch import get_epoch
+from epoch import align_epochs, get_epoch
 from file_helper import generate_file
+from grpc_responses import OBJECT_NOT_FOUND
 from http_gate import (
     attr_into_str_header_curl,
     get_object_and_verify_hashes,
     try_to_get_object_and_expect_error,
     upload_via_http_gate_curl,
 )
-from python_keywords.neofs_verbs import get_netmap_netinfo, head_object
+from python_keywords.neofs_verbs import get_netmap_netinfo, get_object_from_random_node, head_object
 from wellknown_acl import PUBLIC_ACL
 
 from steps.cluster_test_base import ClusterTestBase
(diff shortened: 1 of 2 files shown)


abereziny:
If this is experimental fix, we should not merge it until it's proven to help

robot/resources/lib/python_keywords/epoch.py
```diff
@@ -33,15 +34,28 @@ def ensure_fresh_epoch(
...
+    unique_epochs = list(set(epochs))
+    assert (
+        len(unique_epochs) == 1
+    ), f"unaligned epochs found,  {epochs}, count of unique epochs {len(unique_epochs)}"
+
+
 @allure.step("Get Epoch")
 def get_epoch(shell: Shell, cluster: Cluster, alive_node: Optional[StorageNode] = None):
     alive_node = alive_node if alive_node else cluster.storage_nodes[0]
+    endpoint = alive_node.get_rpc_endpoint() if alive_node else cluster.default_rpc_endpoint
```
abereziny, line 52: why? alive_node already have `if else`

robot/resources/lib/python_keywords/epoch.py
```diff
@@ -33,15 +34,28 @@ def ensure_fresh_epoch(
     return epoch
 
 
+@allure.step("Align epochs for the whole cluster")
+@wait_for_success(60, 5)
+def align_epochs(shell: Shell, cluster: Cluster) -> bool:
```
abereziny, line 39: Shouldn't we add this to the tick_epoch call?

pytest_tests/testsuites/services/http_gate/test_http_system_header.py
```diff
@@ -368,9 +390,17 @@ def test_http_rfc_object_unavailable_after_expir(
...
             ), f"Epochs should be equal: {get_epoch(self.shell, self.cluster)} != {expected_epoch + 1}"
-            try_to_get_object_and_expect_error(
-                cid=user_container,
-                oid=oid,
-                error_pattern="404 Not Found",
-                endpoint=self.cluster.default_http_gate_endpoint,
-            )
+
+            with allure.step("Check object deleted because it expires-on epoch"):
+                align_epochs(self.shell, self.cluster)
```
abereziny, line 395: if we add this to tick_epoch instead we don't need to check it every time in each test.

robot/resources/lib/python_keywords/epoch.py
vkarak1: I am not sure that it's necessary. I decided to add this check only for the cases when we are dealing with expired objects and epochs should be the same in the whole cluster.

robot/resources/lib/python_keywords/epoch.py
vkarak1: corrected, thanks!

abereziny (approved):

anikeev-yadro (approved):

PR #512, opened 2023-02-07
Add assert_s3_acl

EliChin:
Add `assert_s3_acl` func to refactor ACLs assertion process for S3.

Test runs: https://obj-jenkins.spb.yadro.com/view/Ready%20to%20use/job/sbercloud_test/578/allure/, https://obj-jenkins.spb.yadro.com/view/Ready%20to%20use/job/sbercloud_test/579/allure/

Signed-off-by: Elizaveta Chichindaeva <elizaveta@nspcc.ru>

### DIFF
--- a/pytest_tests/helpers/s3_helper.py
+++ b/pytest_tests/helpers/s3_helper.py
@@ -1,4 +1,5 @@
 import datetime
+import logging
 import os
 from datetime import datetime, timedelta
 from typing import Optional
@@ -7,6 +8,8 @@ import allure
 import s3_gate_bucket
 import s3_gate_object
 from dateutil.parser import parse
+
+logger = logging.getLogger("NeoLogger")
 
 
 @allure.step("Expected all objects are presented in the bucket")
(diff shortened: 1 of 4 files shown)


pytest_tests/helpers/s3_helper.py
```diff
@@ -127,3 +130,27 @@ def assert_object_lock_mode(
...
         ).days == retain_period, f"Expected retention period is {retain_period} days"
+
+
+def assert_s3_acl(acl_grants: dict, permitted_users: str):
+    if permitted_users == "AllUsers":
+        for i in range(len(acl_grants)):
+            if acl_grants[i].get("Grantee", {}).get("URI") is not None:
+                uri = acl_grants[i].get("Grantee", {}).get("URI")
+                permission = acl_grants[i].get("Permission")
+        assert (uri, permission) == (
```
dansingjulia, line 141: check only last grant, not all grands in list.

EliChin: I thought it would be right to check if there is no block (left by any chance\bag) granting full control to all users if we expect only canonical user(-s) to have full control.

pytest_tests/helpers/s3_helper.py
```diff
@@ -127,3 +130,27 @@ def assert_object_lock_mode(
         assert (
             retain_date - last_modify + timedelta(seconds=1)
         ).days == retain_period, f"Expected retention period is {retain_period} days"
+
+
+def assert_s3_acl(acl_grants: dict, permitted_users: str):
+    if permitted_users == "AllUsers":
+        for i in range(len(acl_grants)):
```
abereziny, line 137: Why to add complexity and iterate using `i`?
Can we use this instead?
```
for acl_grant in acl_grants:
     acl_grant.get(...) ... 
    ...
```

pytest_tests/testsuites/services/s3_gate/test_s3_object.py
```diff
@@ -914,11 +895,7 @@ def test_s3_sync_dir(self, sync_type, bucket, simple_object_size):
                     obj_head.get("Metadata") == object_metadata
                 ), f"Metadata of object is {object_metadata}"
                 # obj_acl = s3_gate_object.get_object_acl_s3(self.s3_client, bucket, obj_key)
-                # obj_permission = [permission.get("Permission") for permission in obj_acl]
-                # assert obj_permission == [
-                #     "FULL_CONTROL",
-                #     "FULL_CONTROL",
-                # ], "Permission for all groups is FULL_CONTROL"
+                # assert_s3_acl(acl_grants = obj_acl, permitted_users = "AllUsers")
```
abereziny, line 898: Why is this commented?) Do we need it?

pytest_tests/helpers/s3_helper.py
EliChin: Soon, I'll commit a change to this block.
But the logic here is that not every `acl_grant` contains permission for All Users, others may contain "Canonical User".

abereziny: My comment only relates to style of iteration.
Basically, instead of 
```
for i in range(len(Xs)):
     x[i] <...>
```

It is better to use
```
for x in Xs:
     x <...>
```

Because we are working only with one instance of Xs at a time. 
We don't do something like
```
Xs[i+1] = Xs[i] 
if Xs[i-1] > Xs[i]
```
And we don't need `i` elsewhere either.

pytest_tests/testsuites/services/s3_gate/test_s3_object.py
dansingjulia: Check this assert on actual vertsion.

pytest_tests/helpers/s3_helper.py
EliChin: Done

pytest_tests/testsuites/services/s3_gate/test_s3_object.py
EliChin: Checked, left commented

abereziny (approved):

PR #514, opened 2023-03-14
Update Python to 3.10 version.

vvarg229:
This is required for compatibility with neofs-testlib version 1.1.1.


roman-khimov (approved):
OK for me, but maybe @aprasolova knows something about this.

PR #515, opened 2023-03-14
Add the ability to save configuration files.

vvarg229:
Parsing failed tests requires not only logs, but also various configuration files. Added the ability to save these files in a separate archive.

### DIFF
--- a/.devenv.hosting.yaml
+++ b/.devenv.hosting.yaml
@@ -7,7 +7,7 @@ hosts:
       container_name: s01
       config_path: ../neofs-dev-env/services/storage/.storage.env
       wallet_path: ../neofs-dev-env/services/storage/wallet01.json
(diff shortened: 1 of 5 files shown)


.devenv.hosting.yaml
```diff
@@ -7,7 +7,7 @@ hosts:
       container_name: s01
       config_path: ../neofs-dev-env/services/storage/.storage.env
       wallet_path: ../neofs-dev-env/services/storage/wallet01.json
-      local_config_path: ./TemporaryDir/empty-password.yml
+      local_config_path: ./ConfigsDir/empty-password.yml
```
roman-khimov, line 10: Do we really need these? IIUC it's about wallet unlock files with known passwords, nothing really special/unknown here.

PR #517, opened 2023-03-14
Add allure output to generate_object about storage object to container.

vvarg229:
It helps in parsing failed tests.

### DIFF
--- a/pytest_tests/helpers/container.py
+++ b/pytest_tests/helpers/container.py
@@ -74,14 +74,14 @@ class StorageContainer:
                     bearer=bearer_token,
                     wallet_config=wallet_config,
                 )
-
-            storage_object = StorageObjectInfo(
-                container_id,
-                object_id,
-                size=size,
-                wallet_file_path=wallet_path,
-                file_path=file_path,
-                file_hash=file_hash,
-            )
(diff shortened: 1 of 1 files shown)


pytest_tests/helpers/container.py
```diff
@@ -74,14 +74,14 @@ def generate_object(
...
-
-            storage_object = StorageObjectInfo(
-                container_id,
-                object_id,
-                size=size,
-                wallet_file_path=wallet_path,
-                file_path=file_path,
-                file_hash=file_hash,
-            )
+            with allure.step(f"Storage object with size {size} to container {container_id}"):
```
roman-khimov, line 77: _store_ object of size to?

roman-khimov (commented):
Please also use "component:" prefix for commit messages if possible.

vvarg229:
> Please also use "component:" prefix for commit messages if possible.

Thanks. 
In neofs-testcases project, previous commits were made without a "component:" prefix.
But it seems like a really good idea.

pytest_tests/helpers/container.py
vvarg229: Fixed

PR #518, opened 2023-03-14
Add information about an object before deleting it.

vvarg229:
It helps in parsing failed tests.
This way you can verify that exactly those objects have been deleted.


roman-khimov (approved):
Seems somewhat like a separate tested thing, but OK if helps in debugging.

PR #562, opened 2023-05-21
pytest_tests: Reduce log verbosity

vvarg229:
This commit reduces log verbosity during testing by changing the log level from DEBUG to ERROR in pytest.ini. This offers several benefits:

1. Readability: Test results are easier to read with less log output.

2. Speed: Deploying test results to GitHub Pages' allure-report is faster, as there is less data to process and transfer.

3. Disk space: Less output saves space on the GitHub-hosted runners, which have limited storage.

In conclusion, this commit improves the efficiency and usability of the testing process.

### DIFF
--- a/pytest_tests/pytest.ini
+++ b/pytest_tests/pytest.ini
@@ -1,6 +1,6 @@
 [pytest]
 log_cli = 1
-log_cli_level = DEBUG
+log_cli_level = ERROR
 log_cli_format = %(asctime)s [%(levelname)4s] %(message)s
 log_format = %(asctime)s [%(levelname)4s] %(message)s
 log_cli_date_format = %Y-%m-%d %H:%M:%S


pytest_tests/pytest.ini
```diff
@@ -1,6 +1,6 @@
 [pytest]
 log_cli = 1
-log_cli_level = DEBUG
+log_cli_level = ERROR
```
roman-khimov, line 3: How about INFO of some kind? Is ERROR enough to untangle test failures?

vvarg229: > How about INFO of some kind? Is ERROR enough to untangle test failures?

First, in the report will be the test results:
https://nspcc-dev.github.io/neofs-node/95/

If the test falls, the full error log will be available:
https://github.com/nspcc-dev/neofs-node/actions/runs/5025178563/jobs/9011808006

image

roman-khimov: OK, let's try it this way.

roman-khimov (approved):

PR #569, opened 2023-06-01
Add a script to fix ripemd160

vvarg229:
Hashlib uses OpenSSL for ripemd160 and apparently OpenSSL disabled some older crypto algos around version 3.0 in November 2021. All the functions are still there but require manual enabling. But we use ripemd160 for tests.

For the system tests on GitHub runners to change the OpenSSL configuration files, a script and tests for it have been added.
Of course, the user can also run this script on servers or laptops, but it's main purpose is to prepare the environment on ubuntu-latest runners.

For example, For example, here is a demo build without the fix:
https://github.com/nspcc-dev/neofs-node/actions/runs/5148440999/jobs/9270133272

And here's the build with the fix:
https://github.com/nspcc-dev/neofs-node/actions/runs/5148908407/jobs/9271505417


roman-khimov (approved):

AnnaShaleva (approved):
LGTM!

Sorry for nitpicking, but s/`it's main purpose`/`its main purpose ` in the commit message.

PR #572, opened 2023-06-10
Fix "No space left"

vvarg229:
In this PR I fixed https://github.com/nspcc-dev/neofs-testcases/issues/571 and reorganized the tests a bit.
Each of the commits is independent and contains a description of the changes. But in short, now after each test module the temporary files and objects are cleared, the logs for each test are available in the report, the test marks are updated.

### DIFF
--- a/pytest_tests/testsuites/conftest.py
+++ b/pytest_tests/testsuites/conftest.py
@@ -13,6 +13,8 @@ from binary_version_helper import get_local_binaries_versions, get_remote_binari
 from cluster import Cluster
 from common import (
     ASSETS_DIR,
+    TEST_FILES_DIR,
+    TEST_OBJECTS_DIR,
     COMPLEX_OBJECT_CHUNKS_COUNT,
     COMPLEX_OBJECT_TAIL_SIZE,
     FREE_STORAGE,
@@ -144,6 +146,22 @@ def temp_directory() -> str:
 
     with allure.step("Remove tmp directory"):
         remove_dir(full_path)
+
+
+@pytest.fixture(scope="module", autouse=True)
+@allure.title(f'Prepare test files directories')
+def artifacts_directory(temp_directory: str) -> None:
+    dirs = [TEST_FILES_DIR, TEST_OBJECTS_DIR]
+    for dir_name in dirs:
+        with allure.step(f"Prepare {dir_name} directory"):
+            full_path = os.path.join(temp_directory, dir_name)
+            create_dir(full_path)
+
+    yield
+
+    for dir_name in dirs:
+        with allure.step(f"Remove {dir_name} directory"):
+            remove_dir(full_path)
 
 
 @pytest.fixture(scope="session", autouse=True)
--- a/pytest_tests/testsuites/conftest.py
+++ b/pytest_tests/testsuites/conftest.py
@@ -3,6 +3,7 @@ import os
 import re
 import shutil
 import uuid
+import hashlib
 from datetime import datetime
 
 import allure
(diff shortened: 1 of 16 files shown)


pytest_tests/testsuites/conftest.py
```diff
@@ -134,35 +135,47 @@ def check_binary_versions(request, hosting: Hosting, client_shell: Shell):
...
-    dump_logs(hosting, logs_dir, start_time, end_time)
-    attach_logs(logs_dir)
+    store_logs(hosting, logs_dir, test_name, start_time, end_time)
     check_logs(logs_dir)
 
 
+@pytest.fixture(scope="function", autouse=True)
+@allure.title("Collect test logs")
+def collect_test_logs(request, temp_directory, hosting: Hosting):
+    test_name = request.node.nodeid.replace(":", "_").replace("[", "_").replace("]", "_").replace("/", "_")
```
roman-khimov, line 182: `translate()`?

pytest_tests/testsuites/conftest.py
```diff
@@ -134,35 +135,47 @@ def check_binary_versions(request, hosting: Hosting, client_shell: Shell):
...
+    hash_suffix = hashlib.md5(test_name.encode()).hexdigest()
+    file_name = (test_name[:200] + '_' + hash_suffix)  # limit total length to 255
```
roman-khimov, line 184: Do we need a hash if `test_name` fits into `200`?

pytest_tests/testsuites/services/s3_gate/test_s3_ACL.py
```diff
@@ -50,6 +50,7 @@ def test_s3_object_ACL(self, bucket, simple_object_size):
             assert_s3_acl(acl_grants=obj_acl, permitted_users="AllUsers")
 
     @allure.title("Test S3: Bucket ACL")
+    @pytest.mark.skip(reason="https://github.com/nspcc-dev/neofs-s3-gw/issues/784")
```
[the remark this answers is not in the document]
roman-khimov: Maybe not even relevant now with nspcc-dev/neofs-s3-gw#784 fixed.

pytest_tests/testsuites/conftest.py
vvarg229: > Do we need a hash if `test_name` fits into `200`?

In my opinion, yes, hash is needed in this case too.
Standard use case - comparing two test runs. It's very convenient to download two archives and compare them by hash.
It seems to me that it will make life easier:)

pytest_tests/testsuites/conftest.py
roman-khimov: I mean you can do `translate()` and not `replace().replace().replace().replace()`, it'd be more efficient and readable at the same time.

roman-khimov: > Linux does not allow such characters in file names

BTW, it's not true, all of these symbols are perfectly fine. Inconvenient in some cases, true, but valid:
```
$ touch 'test_container.py::TestContainer::test_container_creation[No name]'
$ ls -l 'test_container.py::TestContainer::test_container_creation[No name]'
-rw-r--r-- 1 rik users 0 июн 14 12:57 test_container.py::TestContainer::test_container_creation[No name]
```

vvarg229: > > Linux does not allow such characters in file names
> 
> BTW, it's not true, all of these symbols are perfectly fine. Inconvenient in some cases, true, but valid:
> 
> ```
> $ touch 'test_container.py::TestContainer::test_container_creation[No name]'
> $ ls -l 'test_container.py::TestContainer::test_container_creation[No name]'
> -rw-r--r-- 1 rik users 0 июн 14 12:57 test_container.py::TestContainer::test_container_creation[No name]
> ```

Yes, but it is parsed inside tests without quotes, and the path is part of the return value.

I changed to translate(), if it critical, I can change the name to hash.
Anyway, in allure report attached files are named differently, it's important only for local runs and parsing test crashes.

roman-khimov (approved):

Issue #579, opened 2023-07-04
Test container removal/setEACL with session issued by a NON-owner of the container


cthulhu-rider:
container MUST NOT be removed/setEACLed by anyone except:
* owner (creator)
* trusted party proved by the container owner via session token

AFAIK we have a test that asserts true-positive case (session signed by owner => removal/setEACL OK)

we must additionally test true-negative case:
* issue session on behalf of any party other than container owner (i.e. sign via non-owner key)
* sent removal/setEACL request with token of this session
* expect operation failure

cthulhu-rider:
recent NeoFS CLI revisions check session ownership, so attempting to use another wallet will fail with `session issuer differs with the container owner`. To skip this check and make behavior similar to owner's sessions, use `--force` flag.

with `--force`:
1. success leads to command exit with code=0. 
2. any failure leads to command exit with code=1 and `timed out after 90 seconds` output message

currently, NeoFS CLI will exit with code=0 in case 2, so right now test may rely on code=1 and/or mentioned text output. This may be changed in the future https://github.com/nspcc-dev/neofs-node/issues/2380

PR #582, opened 2023-07-13
acl, container: Add negative tests

vvarg229:
Added two negative tests that verify that the container cannot be removed or set container EACL by anyone other than the owner (creator) or trusted trusted party proved by the container owner via session token.

### DIFF
--- a/pytest_tests/testsuites/acl/test_eacl.py
+++ b/pytest_tests/testsuites/acl/test_eacl.py
@@ -2,6 +2,7 @@ import allure
 import pytest
 from cluster_test_base import ClusterTestBase
 from failover_utils import wait_object_replication
+from grpc_responses import NOT_CONTAINER_OWNER
 from neofs_testlib.shell import Shell
 from python_keywords.acl import (
     EACLAccess,
@@ -28,6 +29,7 @@ from python_keywords.object_access import (
     can_put_object,
     can_search_object,
 )
+from testsuites.acl.conftest import Wallets
 from wellknown_acl import PUBLIC_ACL
 
 
@@ -35,7 +37,7 @@ from wellknown_acl import PUBLIC_ACL
 @pytest.mark.acl_extended
 class TestEACLContainer(ClusterTestBase):
     @pytest.fixture(scope="function")
-    def eacl_full_placement_container_with_object(self, wallets, file_path) -> str:
+    def eacl_full_placement_container_with_object(self, wallets, file_path) -> tuple[str, str, str]:
         user_wallet = wallets.get_wallet()
         storage_nodes = self.cluster.storage_nodes
         node_count = len(storage_nodes)
@@ -678,4 +680,29 @@ class TestEACLContainer(ClusterTestBase):
                     shell=self.shell,
                     endpoint=endpoint,
                     wallet_config=storage_wallet.config_path,
+                )
+
+    @pytest.mark.trusted_party_proved
+    @allure.title("Not owner and not trusted party can NOT set eacl")
+    def test_only_owner_can_set_eacl(
+        self,
+        wallets: Wallets,
+        eacl_full_placement_container_with_object: tuple[str, str, str],
+        not_owner_wallet: str
+    ):
+        cid, oid, file_path = eacl_full_placement_container_with_object
+
+        eacl = [
+            EACLRule(access=EACLAccess.DENY, role=EACLRole.USER, operation=op)
+            for op in EACLOperation
+        ]
+
+        with allure.step("Try to change EACL"):
+            with pytest.raises(RuntimeError, match=NOT_CONTAINER_OWNER):
+                set_eacl(
+                    wallet_path=not_owner_wallet,
+                    cid=cid,
+                    eacl_table_path=create_eacl(cid, eacl, shell=self.shell),
+                    shell=self.shell,
+                    endpoint=self.cluster.default_rpc_endpoint,
                 )
--- a/pytest_tests/testsuites/conftest.py
+++ b/pytest_tests/testsuites/conftest.py
@@ -9,6 +9,7 @@ from datetime import datetime
 import allure
 import pytest
 import yaml
+from typing import Optional
 from binary_version_helper import get_local_binaries_versions, get_remote_binaries_versions
 from cluster import Cluster
 from common import (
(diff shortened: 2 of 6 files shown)


pytest_tests/testsuites/container/test_container.py
```diff
@@ -86,6 +87,47 @@ def test_container_creation(self, default_wallet, name):
...
+        with allure.step("Try to delete container"):
+            try:
+                delete_container(
+                    wallet=not_owner_wallet,
+                    cid=cid,
+                    shell=self.shell,
+                    endpoint=self.cluster.default_rpc_endpoint,
+                    await_mode=True,
+                )
+            except Exception as e:
```
roman-khimov, line 109: What if no exception occurs? Will the test fail?

vvarg229: > Will the test fail?

Yes, the test will fall

roman-khimov:
These tests are OK (when fixed), but in #579 we wanted to check for session-based actions as well.

pytest_tests/testsuites/container/test_container.py
vvarg229: > What if no exception occurs? Will the test fail?

Fixed

vvarg229:
> These tests are OK (when fixed), but in #579 we wanted to check for session-based actions as well.

OK, I'll add session-based tests

pytest_tests/testsuites/acl/test_eacl.py
```diff
@@ -679,3 +679,33 @@ def test_extended_actions_system(self, wallets, eacl_container_with_objects):
...
+        with allure.step("Try to change EACL"):
+            try:
+                set_eacl(
+                    wallet_path=not_owner_wallet,
+                    cid=cid,
+                    eacl_table_path=create_eacl(cid, eacl, shell=self.shell),
+                    shell=self.shell,
+                    endpoint=self.cluster.default_rpc_endpoint,
+                )
+            except Exception as e:
```
roman-khimov, line 708: Same thing with `error_occurred` here, btw.

vvarg229:
> OK, I'll add session-based tests

Done

roman-khimov (commented):
It's nice, but I've noticed that `generate_session_token()` actually fills in `ownerID` from `owner_wallet` which then signs the token in `sign_session_token()`. What also need is a _correct_ `ownerID` in the token, but a token that is signed by a different wallet.

vvarg229:
> It's nice, but I've noticed that `generate_session_token()` actually fills in `ownerID` from `owner_wallet` which then signs the token in `sign_session_token()`. What also need is a _correct_ `ownerID` in the token, but a token that is signed by a different wallet.

Thank you for the comment. I'll also add a related case.

vvarg229:
> These tests are OK (when fixed), but in #579 we wanted to check for session-based actions as well.

Done.

I have a question in the process of writing tests: https://github.com/nspcc-dev/neofs-node/issues/2466

vvarg229:
> > These tests are OK (when fixed), but in #579 we wanted to check for session-based actions as well.
> 
> Done.
> 
> I have a question in the process of writing tests: [nspcc-dev/neofs-node#2466](https://github.com/nspcc-dev/neofs-node/issues/2466)

Question resolved

pytest_tests/testsuites/acl/test_eacl.py
```diff
@@ -679,3 +680,29 @@ def test_extended_actions_system(self, wallets, eacl_container_with_objects):
                     endpoint=endpoint,
                     wallet_config=storage_wallet.config_path,
                 )
+
+
+    @pytest.mark.trusted_party_proved
+    @allure.title("Not owner and not trusted party can NOT set eacl")
+    def test_only_owner_can_set_eacl(
+        self,
+        wallets,
```
evgeniiz321, line 689: It is a good thing to use type annotations as much as possible, even when it is some fixtures. We can specify here a type that is returned from fixtures.

pytest_tests/testsuites/conftest.py
```diff
@@ -304,33 +305,21 @@ def background_grpc_load(client_shell: Shell, hosting: Hosting):
...
-            wallet_from_path=wallet_path,
-            wallet_from_password=WALLET_PASS,
-        )
+@pytest.fixture(scope="function")
+@allure.title("Prepare not owner wallet and deposit")
+def not_owner_wallet(client_shell: Shell, temp_directory: str, cluster: Cluster):
+    wallet_path = create_wallet(client_shell, temp_directory, cluster, 'not_owner_wallet')
+    try:
+        yield wallet_path
+    finally:
```
evgeniiz321, line 314: as far as I remember, there is no need to handle exceptions in fixtures like these, in case of errors inside other fixtures/tests - the teardown code will still be invoked https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#handling-errors-for-yield-fixture

pytest_tests/testsuites/acl/test_eacl.py
```diff
@@ -679,3 +680,29 @@ def test_extended_actions_system(self, wallets, eacl_container_with_objects):
...
+        not_owner_wallet
+    ):
+        cid, oid, file_path = eacl_full_placement_container_with_object
+
+        eacl = [
+            EACLRule(access=EACLAccess.DENY, role=EACLRole.USER, operation=op)
+            for op in EACLOperation
+        ]
+
+        with allure.step("Try to change EACL"):
```
roman-khimov, line 700: Try forced set as well?

pytest_tests/testsuites/conftest.py
vvarg229: > as far as I remember, there is no need to handle exceptions in fixtures like this, in case of errors inside other fixtures/tests - the teardown code will still be invoked https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#handling-errors-for-yield-fixture

Fixed

pytest_tests/testsuites/acl/test_eacl.py
vvarg229: > It is a good thing to use type annotations as much as possible, even when it is some fixtures. We can specify here a type that is returned from fixtures.

Fixed

pytest_tests/testsuites/acl/test_eacl.py
vvarg229: > Try forced set as well?

Does container set-eacl have a forced option?
```
$ neofs-cli container set-eacl --help
Set new extended ACL table for container.
Container ID in EACL table will be substituted with ID from the CLI.

Usage:
  neofs-cli container set-eacl [flags]

Flags:
      --address string        Address of wallet account
      --await                 block execution until EACL is persisted
      --cid string            Container ID.
  -g, --generate-key          Generate new private key
  -h, --help                  help for set-eacl
      --no-precheck           do not pre-check the extensibility of the container ACL
  -r, --rpc-endpoint string   Remote node address (as 'multiaddr' or '<host>:<port>')
      --session string        Filepath to a JSON- or binary-encoded token of the container SETEACL session
      --table string          path to file with JSON or binary encoded EACL table
  -t, --timeout duration      Timeout for the operation (default 15s)
      --ttl uint32            TTL value in request meta header (default 2)
  -w, --wallet string         Path to the wallet
  -x, --xhdr strings          Request X-Headers in form of Key=Value

Global Flags:
  -c, --config string   Config file (default is $HOME/.config/neofs-cli/config.yaml)
  -v, --verbose         Verbose output
```

vvarg229:
After the PR https://github.com/nspcc-dev/neofs-node/pull/2501 the error messages have changed.
I changed it and in the test, here are the new results:
https://http.t5.fs.neo.org/86C4P6uJC7gb5n3KkwEGpXRfdczubXyRNW5N9KeJRW73/323-1692901993/index.html#

roman-khimov (approved):

evgeniiz321 (approved):

PR #592, opened 2023-08-07
Add a script to collect logs from containers

vvarg229:
Added a script for collecting logs from docker containers and merging them into an archive.
This is needed to collect logs and debug in case tests do not run, for example, if dev-env on CI does not run.

Note - this script only collects logs to the archive, sending logs to NeoFS is done on the github actions side.

### DIFF
--- a/tools/src/zip_dev_env_logs.py
+++ b/tools/src/zip_dev_env_logs.py
@@ -1,0 +1,48 @@
+import docker
+import logging
+import os
+import zipfile
+import tempfile
+
+
+def setup_logging():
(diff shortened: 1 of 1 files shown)


roman-khimov:
Is it gonna be executed as an additional step in GH Action? Only on failures or always?

vvarg229:
> Is it gonna be executed as an additional step in GH Action? Only on failures or always?
It will be separate steps, here's the PR that has the necessary changes:

https://github.com/nspcc-dev/neofs-node/pull/2476

roman-khimov (approved):

PR #598, opened 2023-08-14
Fix logging level in the assert_s3_acl

vvarg229:
Logging level changed from error to warning.
Now there will be no such messages in the logs:

-------------------------------- live log call --------------------------------- 
Error: -14 10:10:37 [ERROR] FULL_CONTROL is given to All Users
PASSED                                                                   [ 62%]

### DIFF
--- a/pytest_tests/helpers/s3_helper.py
+++ b/pytest_tests/helpers/s3_helper.py
@@ -157,4 +157,4 @@ def assert_s3_acl(acl_grants: list, permitted_users: str):
                 permission = acl_grant.get("Permission")
                 assert permission == "FULL_CONTROL", "Only CanonicalUser should have FULL_CONTROL"
             else:
-                logger.error("FULL_CONTROL is given to All Users")
+                raise AssertionError("FULL_CONTROL is given to All Users")
--- a/pytest_tests/testsuites/services/s3_gate/test_s3_object.py
+++ b/pytest_tests/testsuites/services/s3_gate/test_s3_object.py
@@ -133,7 +133,7 @@ class TestS3GateObject(TestS3GateBase):
 
         with allure.step("Copy object and check acl attribute"):
             copy_obj_path = s3_gate_object.copy_object_s3(
-                self.s3_client, bucket, obj_key, ACL="public-read-write"
+                self.s3_client, bucket, obj_key, ACL="private"
             )
             obj_acl = s3_gate_object.get_object_acl_s3(self.s3_client, bucket, copy_obj_path)
             assert_s3_acl(acl_grants=obj_acl, permitted_users="CanonicalUser")


roman-khimov:
Are you sure it's a logger problem? It seems more like an S3 GW issue to me and a test issue as well (it only logs, but should fail).

roman-khimov:
@smallhive, can you check what's up there? This seems to be very suspicious to me.

Refs. nspcc-dev/neofs-s3-gw#801.

smallhive:
> @smallhive, can you check what's up there? This seems to be very suspicious to me.
> 
> Refs. [nspcc-dev/neofs-s3-gw#801](https://github.com/nspcc-dev/neofs-s3-gw/issues/801).

actions-runner/_work/neofs-node/neofs-node/neofs-testcases/pytest_tests/helpers/s3_helper.py

```python
def assert_s3_acl(acl_grants: list, permitted_users: str):
    ...

    if permitted_users == "CanonicalUser":
        for acl_grant in acl_grants:
            if acl_grant.get("Grantee", {}).get("Type") == "CanonicalUser":
                ...
            else:
                logger.error("FULL_CONTROL is given to All Users")
```

roman-khimov:
I'm more about the actual gw behavior, logs are secondary to it.

smallhive:
After deeper investigation and with @vvarg229 discussion we decided - it is a problem, but with test.
```python
    def test_s3_copy_acl(self, bucket, simple_object_size):
        ...
        with allure.step("Copy object and check acl attribute"):
            copy_obj_path = s3_gate_object.copy_object_s3(
                self.s3_client, bucket, obj_key, ACL="public-read-write"
            )
            obj_acl = s3_gate_object.get_object_acl_s3(self.s3_client, bucket, copy_obj_path)
            assert_s3_acl(acl_grants=obj_acl, permitted_users="CanonicalUser")
```

We are copying objects but set ACL="public-read-write", which means Full control for all users. According to [the gate code](https://github.com/nspcc-dev/neofs-s3-gw/blob/master/api/handler/acl.go#L585) in this case, we give access to everyone.

To fix the test and remove _FULL_CONTROL is given to All Users_ we need to set ACL="private".
I tend to think it was an error from a previous commit. It was feractor

vvarg229:
> After deeper investigation and with @vvarg229 discussion we decided - it is a problem, but with test.
> 
> ```python
>     def test_s3_copy_acl(self, bucket, simple_object_size):
>         ...
>         with allure.step("Copy object and check acl attribute"):
>             copy_obj_path = s3_gate_object.copy_object_s3(
>                 self.s3_client, bucket, obj_key, ACL="public-read-write"
>             )
>             obj_acl = s3_gate_object.get_object_acl_s3(self.s3_client, bucket, copy_obj_path)
>             assert_s3_acl(acl_grants=obj_acl, permitted_users="CanonicalUser")
> ```
> 
> We are copying objects but set ACL="public-read-write", which means Full control for all users. According to [the gate code](https://github.com/nspcc-dev/neofs-s3-gw/blob/master/api/handler/acl.go#L585) in this case, we give access to everyone.
> 
> To fix the test and remove _FULL_CONTROL is given to All Users_ we need to set ACL="private". I tend to think it was an error from a previous commit. It was feractor

Fixed

roman-khimov (approved):
1. I'd reorder these commits to fix the test first and then make it fail in case of wrong ACL, just to have it always pass.
2. `logger.error()` doesn't seem to be useful if we raise an error anyway.
3. These fixes should not sit in the queue for a long time, they're trivial.

vvarg229:
> 1. I'd reorder these commits to fix the test first and then make it fail in case of wrong ACL, just to have it always pass.
> 2. `logger.error()` doesn't seem to be useful if we raise an error anyway.
> 3. These fixes should not sit in the queue for a long time, they're trivial.

Done

PR #601, opened 2023-08-15
Add expire argument to process-allure-reports.py

vvarg229:
These changes will be used here:
https://github.com/nspcc-dev/neofs-node/pull/2511

### DIFF
--- a/tools/src/process-allure-reports.py
+++ b/tools/src/process-allure-reports.py
@@ -17,16 +17,23 @@ def parse_args():
     parser.add_argument('--cid', required=True, type=str, help='Container ID')
     parser.add_argument('--run_id', required=True, type=str, help='GitHub run ID')
     parser.add_argument('--allure_report', type=str, help='Path to generated allure report directory',
-                        default='allure_report')
+                        default='allure_report'),
+    parser.add_argument('--expire-at', type=int,
+                        help='Expiration epoch. If epoch is not provided, or if it is 0, the report will be stored indefinitely',
+                        default=None)
+
     return parser.parse_args()
 
 
 def put_combine_result_as_static_page(directory: str, neofs_domain: str, wallet: str, cid: str, run_id: str,
-                                      password: str) -> None:
+                                      expire_at: int, password: str) -> None:
     base_cmd = (
         f'NEOFS_CLI_PASSWORD={password} neofs-cli --rpc-endpoint st1.{neofs_domain}:8080 '
         f'--wallet {wallet}  object put --cid {cid} --timeout {PUT_TIMEOUT}s'
     )
+
+    if expire_at is not None and expire_at > 0:
+        base_cmd += f' --expire-at {expire_at}'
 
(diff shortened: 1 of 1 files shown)


tools/src/process-allure-reports.py
```diff
@@ -17,17 +17,25 @@ def parse_args():
     parser.add_argument('--cid', required=True, type=str, help='Container ID')
     parser.add_argument('--run_id', required=True, type=str, help='GitHub run ID')
     parser.add_argument('--allure_report', type=str, help='Path to generated allure report directory',
-                        default='allure_report')
+                        default='allure_report'),
+    parser.add_argument('--expire', type=int,
```
carpawell, line 21: so basically that is `--expire_after`? also, you are able to make it in hours, not epochs: get epoch duration and block longevity with `neofs-cli`. not sure if it is required

vvarg229: > make it in hours, not epochs

Can you please give me an example?

carpawell: epoch equals an hour cause it is `Time per block` * `Epoch duration` seconds:

```
▶ neofs-cli netmap netinfo -r grpcs://st1.storage.fs.neo.org:8082                                        
...
Time per block: 15s
NeoFS network configuration (system)
...
  Epoch duration: 240
...
```

that info is always true and available. it can prevent from `(hours for current testnet and mainnet)` comments to script args (if we need it of course)

vvarg229: > epoch equals an hour cause it is `Time per block` * `Epoch duration` seconds:
> 
> ```
> ▶ neofs-cli netmap netinfo -r grpcs://st1.storage.fs.neo.org:8082                                        
> ...
> Time per block: 15s
> NeoFS network configuration (system)
> ...
>   Epoch duration: 240
> ...
> ```
> 
> that info is always true and available. it can prevent `(hours for current testnet and mainnet)` comments to script args (if we need it of course)

Ok, I'll fix the description of the argument.
But anyway, I think it's better to send the epoch to the script.
I calculate the expiration time in the workflow file:
https://github.com/nspcc-dev/neofs-node/pull/2511

roman-khimov: Renaming to `--expires-at` or something (see nspcc-dev/neofs-node#1574) can help. But epochs are OK here, hour->epoch mapping likely can more easily be done by a layer above.

vvarg229: > Renaming to `--expires-at` or something (see [nspcc-dev/neofs-node#1574](https://github.com/nspcc-dev/neofs-node/issues/1574)) can help. But epochs are OK here, hour->epoch mapping likely can more easily be done by a layer above.

Done

carpawell: > @carpawell, are we OK with expires-at here?

totally

roman-khimov (approved):

Issue #556, opened 2023-04-22
Test metabase resync for lock objects


roman-khimov:
The scenario from https://github.com/nspcc-dev/neofs-node/issues/1502. It should be fixed now, but needs to be checked and automated.

roman-khimov:
Merged with #608.

Issue #550, opened 2023-04-07
Lock object tests


roman-khimov:
By @carpawell:

https://github.com/nspcc-dev/neofs-node/pull/1461, v0.30.0

* It is impossible to delete a locked object (control service is an exception).
* It is impossible to delete a lock object.
* It is prohibited to put a lock object without expiration.
* After lock object expiration, locked objects must become free to delete.

vvarg229:
@roman-khimov @carpawell please explain the concept of "control service":
`It is impossible to delete a locked object (control service is an exception).
`

carpawell:
@vvarg229, in short -- `neofs-cli control`. It allows controlling SN/IR. You should be [authorized](https://github.com/nspcc-dev/neofs-node/blob/27623c50edbb86a09e9b4b368be3e3d399a92969/config/example/node.yaml#L64-L67) to be allowed to make control operations. One of the operations is `neofs-cli control drop-objects`. It means "admin want some object to be removed" and that should be done despite any restriction (e.g. our beloved `LOCK` object).

evgeniiz321:
@roman-khimov could you please take a look at current object lock tests - https://http.fs.neo.org/HXSaMJXk2g8C14ht8HSi7BBaiYZ1HeWh2xnWPGQCg4H6/654-1693496239/index.html#suites/b34feeefa7bd0ee424f598738938f10a/cc81bd598bed3fbb/ ? Looks like all common and useful scenarios are covered

roman-khimov:
@carpawell?

carpawell:
Kinda OK. Mb also add "Lock objects can be dropped"? Not sure what exactly is expected then (should their targets be unlocked?) but still.

Also, some suggestions: I was confused a little by some test naming: some of them have `complex`/`simple` in their title but some of them have that difference as a param; control service testing is mentioned by "drop" word and not by some explicit wording in the title. Also, some CLI commands have double spacing (e.g. `neofs-cli control drop-objects  -o`), which is cosmetic of course.

PR #608, opened 2023-08-21
Lock resync

vvarg229:


pytest_tests/testsuites/object/test_object_lock.py
```diff
@@ -663,3 +663,29 @@ def test_link_object_of_complex_object_should_also_be_protected_from_deletion(
...
+        indirect=True,
+    )
+    def test_the_object_lock_should_be_kept_after_restarting_the_nodes(
+            self,
+            locked_storage_object: StorageObjectInfo,
+    ):
+        """
+        Lock objects should fill metabase on resync_metabase
+        """
+        restart_storage_nodes(self.cluster.storage_nodes)
```
roman-khimov, line 681: This is not enough for the nspcc-dev/neofs-node#1502 case, metabase must be resynchronized.

roman-khimov (changes requested):
Do we have other cases from #550 covered?

vvarg229:
> Do we have other cases from #550 covered?

No, the other tests will be added in the next PR.

pytest_tests/testsuites/object/test_object_lock.py
vvarg229: > This is not enough for the [nspcc-dev/neofs-node#1502](https://github.com/nspcc-dev/neofs-node/issues/1502) case, metabase must be resynchronized.

@roman-khimov @carpawell  can you please recommend what is the way to make the metabase resynchronized?

carpawell: I strongly recommend dropping metabase too.

roman-khimov: And there is a `delete_storage_node_data()` for it already in the testlib.

roman-khimov:
BTW, the test you have here is for #556, not #550.

pytest_tests/testsuites/object/test_object_lock.py
vvarg229: @roman-khimov @carpawell If I physically delete the metabase files, should I set resync_metabase to true?
Or should I change the configuration file in the dev-env https://github.com/nspcc-dev/neofs-dev-env/blob/master/services/storage/cfg/config.yml#L39 ?
Why is resync_metabase false?

roman-khimov: > If I physically delete the metabase files, should I set resync_metabase to true?

Yes.

> Why is resync_metabase false?

It's never `true` in production, this doesn't make much sense, it's only needed when there are some problems with it (storage failure or meta-incompatible node upgrade). So dev-env should have it set to `false` as usual, it's only some specific test that wants to do something. BTW, maybe it'd be easier for you to use environment variable to override this setting.

vvarg229: After deleting the metabase files from all node object deletion fails with an error:
`gRPC dial: context deadline exceeded`

https://http.t5.fs.neo.org/86C4P6uJC7gb5n3KkwEGpXRfdczubXyRNW5N9KeJRW73/334-1694095194/index.html#suites/b34feeefa7bd0ee424f598738938f10a/afc0e4a3ea957e70/

@roman-khimov @carpawell please tell me, is this expected behavior?

vvarg229:
> Needs to be rebased

Fixed

robot/variables/common.py
```diff
@@ -25,11 +25,14 @@
...
 DEVENV_PATH = os.getenv("DEVENV_PATH", os.path.join("..", "neofs-dev-env"))
+DOCKER_COMPOSE_CONFIG_FILE = os.getenv("DOCKER_COMPOSE_CONFIG_FILE", os.path.join(DEVENV_PATH, "services", "storage",
+                                                                                  "docker-compose.yml"))
+DOCKER_COMPOSE_ENV_FILE = os.getenv("DOCKER_COMPOSE_ENV_FILE", os.path.join(DEVENV_PATH, "services", "storage",
+                                                                            ".int_test.env"))
 
 # Password of wallet owned by user on behalf of whom we are running tests
 WALLET_PASS = os.getenv("WALLET_PASS", "")
 
-
```
roman-khimov, line 32: :eyes:

pytest_tests/testsuites/object/test_object_lock.py
```diff
@@ -119,6 +123,15 @@ def locked_storage_object(
 
 @pytest.mark.grpc_object_lock
 class TestObjectLockWithGrpc(ClusterTestBase):
+
+    @allure.title("Docker compose restart containers with new env file")
```
roman-khimov, line 126: It restarts the whole cluster, right?

pytest_tests/testsuites/object/test_object_lock.py
```diff
@@ -684,3 +752,45 @@ def test_link_object_of_complex_object_should_also_be_protected_from_deletion(
...
+    )
+    def test_the_object_lock_should_be_kept_after_restarting_the_nodes(
+            self,
+            new_locked_storage_object: StorageObjectInfo,
+            enable_metabase_resync_on_start,
+    ):
+        """
+        Lock objects should fill metabase on resync_metabase
+        """
+        with allure.step(f"Delete metabase files from storage nodes"):
```
roman-khimov, line 771: IIUC this should be done after the initial deletion attempt. Likely that's your problem with `deadline exceeded`.

pytest_tests/testsuites/object/test_object_lock.py
```diff
@@ -684,3 +752,45 @@ def test_link_object_of_complex_object_should_also_be_protected_from_deletion(
...
+            for node in self.cluster.storage_nodes:
+                delete_node_data(node)
```
roman-khimov, line 773: But does it remove metabase only? Looks like it drops everything and that's a radically different thing. The node doesn't have any data at all in this case and there is nothing to resync the metabase from.

pytest_tests/testsuites/object/test_object_lock.py
```diff
@@ -119,6 +123,15 @@ def locked_storage_object(
 
 @pytest.mark.grpc_object_lock
 class TestObjectLockWithGrpc(ClusterTestBase):
+
```
carpawell, line 125: do we need an empty line here? seems like it was not here before

pytest_tests/testsuites/object/test_object_lock.py
```diff
@@ -144,6 +157,61 @@ def new_locked_storage_object(
...
+        replacements = {
+            'NEOFS_STORAGE_SHARD_0_RESYNC_METABASE=false': 'NEOFS_STORAGE_SHARD_0_RESYNC_METABASE=true\n',
+            'NEOFS_STORAGE_SHARD_1_RESYNC_METABASE=false': 'NEOFS_STORAGE_SHARD_1_RESYNC_METABASE=true\n'
+        }
+
+        unprocessed_lines = set(replacements.values())
+
+        modified_lines = []
+
+        for line in lines:
```
carpawell, line 191: well, since that is a tricky hack, isn't appending the file with the required envs an easier approach? or renaming the original file, reading it, creating a new file (and appending the needed lines to it), and then renaming it back?

but that is just a reading simplification. mb you think differently

pytest_tests/testsuites/object/test_object_lock.py
```diff
@@ -684,3 +752,45 @@ def test_link_object_of_complex_object_should_also_be_protected_from_deletion(
...
                 )
+
+    @allure.title("Locked object should be protected from deletion after the storage nodes are restarted")
+    @pytest.mark.parametrize(
+        # Only complex objects are required for this test
+        "new_locked_storage_object",
+        [pytest.lazy_fixture("complex_object_size")],
+        indirect=True,
+    )
+    def test_the_object_lock_should_be_kept_after_restarting_the_nodes(
```
carpawell, line 763: it is not about the lock objects mainly, it is about relations b/w the objects: e.g. lock**ed** object should be still locked. both the lock and the lock**ed** object are likely to be in the storage but a relationship b/w them may be lost

carpawell: the comment is just about the test naming

pytest_tests/testsuites/object/test_object_lock.py
```diff
@@ -684,3 +752,45 @@ def test_link_object_of_complex_object_should_also_be_protected_from_deletion(
                     self.shell,
                     self.cluster.default_rpc_endpoint,
                 )
+
+    @allure.title("Locked object should be protected from deletion after the storage nodes are restarted")
```
carpawell, line 756: not just "restarted" in fact

pytest_tests/testsuites/object/test_object_lock.py
```diff
@@ -684,3 +752,45 @@ def test_link_object_of_complex_object_should_also_be_protected_from_deletion(
...
+    @pytest.mark.parametrize(
+        # Only complex objects are required for this test
```
carpawell, line 758: why?

pytest_tests/testsuites/object/test_object_lock.py
```diff
@@ -2,7 +2,6 @@
 import os
 import re
 import subprocess
-from time import sleep
```
carpawell: why is it dropped in this commit?

carpawell (changes requested):

pytest_tests/testsuites/object/test_object_lock.py
vvarg229: > well, since that is a tricky hack, isn't appending the file with the required envs an easier approach? or renaming the original file, reading it, creating a new file (and appending the needed lines to it), and then renaming it back?
> 
> but that is just a reading simplification. mb you think differently

Yeah, I've been thinking about it.
However, it turns out that I have to copy the docker-compose.yml file from dev-env and change the path to the new file with new environment variables.
It seems to me that this will complicate the code too much and make it incomprehensible.

pytest_tests/testsuites/object/test_object_lock.py
vvarg229: > why?

Is it important to check the simple object here as well? If you think it's important, I'll add simple_object_size too.

pytest_tests/testsuites/object/test_object_lock.py
vvarg229: > It restarts the whole cluster, right?

I fixed the feature name and description - only storage nodes are restarted

pytest_tests/testsuites/object/test_object_lock.py
vvarg229: > do we need an empty line here? seems like it was not here before

Fixed

pytest_tests/testsuites/object/test_object_lock.py
vvarg229: > it is not about the lock objects mainly, it is about relations b/w the objects: e.g. lock**ed** object should be still locked. both the lock and the lock**ed** object are likely to be in the storage but a relationship b/w them may be lost

Fixed

pytest_tests/testsuites/object/test_object_lock.py
vvarg229: > not just "restarted" in fact

Fixed

pytest_tests/testsuites/object/test_object_lock.py
vvarg229: > why is it dropped in this commit?

Fixed - it was a rebase error

robot/variables/common.py
vvarg229: Fixed

pytest_tests/testsuites/object/test_object_lock.py
vvarg229: > IIUC this should be done after the initial deletion attempt. Likely that's your problem with `deadline exceeded`.

I added an attempt to delete the object before deleting the metabase.
The result is the same:
https://http.t5.fs.neo.org/86C4P6uJC7gb5n3KkwEGpXRfdczubXyRNW5N9KeJRW73/339-1694555644/index.html#suites/b34feeefa7bd0ee424f598738938f10a/3a36ea27b17b99e8/

pytest_tests/testsuites/object/test_object_lock.py
vvarg229: > the logic behind the complex objects are complicated enought to treat them an absolutelly different way of object storage, so i would expect testing both always. i have fixed both the bugs that relate the simple objects only and the bugs that affects chains of the objects

Fixed

pytest_tests/testsuites/object/test_object_lock.py
vvarg229: > But does it remove metabase only? Looks like it drops everything and that's a radically different thing. The node doesn't have any data at all in this case and there is nothing to resync the metabase from.

Fixed

pytest_tests/testsuites/object/test_object_lock.py
vvarg229: @roman-khimov @carpawell  please take a look at the test result.
https://http.t5.fs.neo.org/86C4P6uJC7gb5n3KkwEGpXRfdczubXyRNW5N9KeJRW73/339-1694717358/index.html#suites/b34feeefa7bd0ee424f598738938f10a/b60696c863dd0db8/

It seems that I'm deleting the metabase correctly - removing storage/meta0 and meta1:
https://github.com/nspcc-dev/neofs-testlib/blob/master/src/neofs_testlib/hosting/docker_host.py#L137

Is it possible that this is a bug?

pytest_tests/testsuites/object/test_object_lock.py
carpawell: this is a too wide error, usually, it means that a node is not ready to handle connections (not started successfully?)

pytest_tests/testsuites/object/test_object_lock.py
vvarg229: Fixed

pytest_tests/testsuites/object/test_object_lock.py
vvarg229: Fixed

vvarg229:
Colleagues, thank you very much for noticing and participating in the discussion.
All errors have been fixed, here are the test results:
https://http.t5.fs.neo.org/86C4P6uJC7gb5n3KkwEGpXRfdczubXyRNW5N9KeJRW73/351-1695662017/index.html

In order for the tests to work correctly, we need to merge these 2 PRs first:
https://github.com/nspcc-dev/neofs-node/pull/2592
https://github.com/nspcc-dev/neofs-dev-env/pull/291

@carpawell apologies to you - we made changes to the .env files before your review, if we had waited we wouldn't have had to revert those changes now.

robot/resources/lib/python_keywords/node_management.py
```diff
@@ -178,6 +178,17 @@ def delete_node_data(node: StorageNode) -> None:
...
+def delete_node_metadata(node: StorageNode) -> None:
+    """
+    The function deletes metadata from host for specified node.
+    Args:
+        node: node for which metadata should be deleted.
+    """
+    node.stop_service()
+    node.host.delete_storage_node_data(node.name, cache_only=True)
+    time.sleep(parse_time(MORPH_BLOCK_TIME))
+    time.sleep(parse_time(METABASE_RESYNC_TIMEOUT))
```
roman-khimov, line 191: The node is stopped, why waiting here?

vvarg229: > The node is stopped, why waiting here?

Fixed

pytest_tests/testsuites/object/test_object_lock.py
```diff
@@ -684,3 +685,85 @@ def test_link_object_of_complex_object_should_also_be_protected_from_deletion(
...
+            with pytest.raises(Exception, match=OBJECT_IS_LOCKED):
+                delete_object(
+                    new_locked_storage_object.wallet_file_path,
+                    new_locked_storage_object.cid,
+                    new_locked_storage_object.oid,
+                    self.shell,
+                    self.cluster.default_rpc_endpoint,
+                )
+
+        with allure.step("Restart storage nodes to enable resync_metabase"):
```
roman-khimov, line 758: Is it enabled or disabled here? I thought it's enabled via `enable_metabase_resync_on_start`, so it's active when we `start_storage_nodes(self.cluster.storage_nodes)` above and I'd probably expect it to be disabled then on the node restart (back to normal).

roman-khimov (commented):
But OK otherwise. @carpawell, @evgeniiz321?

robot/resources/lib/python_keywords/failover_utils.py
```diff
@@ -71,3 +78,84 @@ def get_morph_chain_endpoints(hosting: Hosting) -> List[Tuple[str, str]]:
         port = str(parsed_url.port)
         endpoints.append((addr, port))
     return endpoints
+
+
```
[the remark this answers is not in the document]
vvarg229: > Too many newlines for me.

https://peps.python.org/pep-0008/#blank-lines
`Surround top-level function and class definitions with two blank lines.
`

In formatting, I do what the IDE tells me to do:

image

pytest_tests/testsuites/object/test_object_lock.py
vvarg229: > Is it enabled or disabled here? I thought it's enabled via `enable_metabase_resync_on_start`, so it's active when we `start_storage_nodes(self.cluster.storage_nodes)` above and I'd probably expect it to be disabled then on the node restart (back to normal).

Oh, I'm sorry, please. It was a debug, I forgot to delete it.
It's fixed now.

robot/resources/lib/python_keywords/failover_utils.py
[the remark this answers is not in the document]
roman-khimov: Meh, OK.

evgeniiz321 (approved):

roman-khimov (approved):

PR #620, opened 2023-08-30
pyproject: drop neofs-keywords

roman-khimov:
It doesn't seem to be used, but I'm not sure what's gonna happen if we merge it. Opinions?


vvarg229 (approved):
Looks good:
https://http.t5.fs.neo.org/86C4P6uJC7gb5n3KkwEGpXRfdczubXyRNW5N9KeJRW73/328-1693473422/index.html

Issue #616, opened 2023-08-28
Tests set EACLs for SYSTEM role, but it breaks the container


roman-khimov:
## Current Behavior
We have some test doing
```
COMMAND: neofs-cli --config /home/runner/work/neofs-node/neofs-node/neofs-testcases/wallet_config.yml acl extended create --cid '9LM9qbY1va3yaHEfkQRkfh4996vqUTgYQorXLi6r3bFq' --out '/home/runner/work/neofs-node/neofs-node/neofs-testcases/TemporaryDir/TestFilesDir/eacl_table_e1f65d81-93eb-470e-8e25-af748cee6e16.json' --rule 'deny put  others' --rule 'deny put  user' --rule 'deny put  system' --rule 'deny get  others' --rule 'deny get  user' --rule 'deny get  system' --rule 'deny head  others' --rule 'deny head  user' --rule 'deny head  system' --rule 'deny getrange  others' --rule 'deny getrange  user' --rule 'deny getrange  system' --rule 'deny getrangehash  others' --rule 'deny getrangehash  user' --rule 'deny getrangehash  system' --rule 'deny search  others' --rule 'deny search  user' --rule 'deny search  system' --rule 'deny delete  others' --rule 'deny delete  user' --rule 'deny delete  system'
```

## Expected Behavior
Do not mess with `system`. It leads to a broken container and nothing else.

## Possible Solution
Drop `system` rules.

## Context
nspcc-dev/neofs-node#2531

roman-khimov:
Fixed?

PR #627, opened 2023-09-05
Deprecate System role in EACL #616

vvarg229:
Removed the System role in EACL and removed the tests related to this role. The System EACL role leads to a broken container and nothing else and will be removed in the issue https://github.com/nspcc-dev/neofs-node/issues/2531

Tests result:
https://http.t5.fs.neo.org/86C4P6uJC7gb5n3KkwEGpXRfdczubXyRNW5N9KeJRW73/332-1693924040/index.html#

### DIFF
--- a/pytest_tests/testsuites/acl/test_eacl.py
+++ b/pytest_tests/testsuites/acl/test_eacl.py
@@ -2,7 +2,7 @@ import allure
 import pytest
 from cluster_test_base import ClusterTestBase
 from failover_utils import wait_object_replication
-from grpc_responses import NOT_CONTAINER_OWNER
+from grpc_responses import NOT_CONTAINER_OWNER, EACL_PROHIBITED_TO_MODIFY_SYSTEM_ACCESS
 from neofs_testlib.shell import Shell
 from python_keywords.acl import (
     EACLAccess,
@@ -11,7 +11,7 @@ from python_keywords.acl import (
     EACLRule,
     create_eacl,
     set_eacl,
-    wait_for_cache_expired,
+    wait_for_cache_expired, get_eacl,
 )
 from python_keywords.container import create_container
 from python_keywords.container_access import (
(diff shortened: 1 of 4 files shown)


carpawell (commented):
Will we add some tests about _future_ system changes? I mean refusing system changes will be added in the `neofs-cli` and IR. Also, some `--force` may be added. So some issues about that?

vvarg229:
> * we just deprecate the role, we should continue to test it
These tests are the cause of cluster breakage, which affects other tests.

> * we forbid it, then we should test that it's denied by the system (try and expect failure)
I agree, it's a good idea to add tests like this.
@roman-khimov what do you think?

vvarg229:
> Will we add some tests about _future_ system changes?
I think we should add tests for not being able to change System role using EACL. But it seems to me that this should be done after the issue https://github.com/nspcc-dev/neofs-node/issues/2531 is resolved.

> I mean refusing system changes will be added in the `neofs-cli` and IR. Also, some `--force` may be added. So some issues about that?
Let's discuss this in the https://github.com/nspcc-dev/neofs-node/issues/2531 issue?

pytest_tests/testsuites/acl/test_eacl.py
```diff
@@ -252,436 +252,6 @@ def test_extended_acl_deny_replication(
                 storage_nodes,
             )
 
-    @allure.title("Testcase to validate NeoFS system operations with extended ACL")
```
roman-khimov, line 255: I don't see anything similar for basic ACL, is that the case?

pytest_tests/testsuites/acl/test_eacl.py
```diff
@@ -252,436 +252,6 @@ def test_extended_acl_deny_replication(
...
-    def test_extended_actions_system(self, wallets, eacl_container_with_objects):
-        user_wallet = wallets.get_wallet()
-        ir_wallet, storage_wallet = wallets.get_wallets_list(role=EACLRole.SYSTEM)[:2]
```
roman-khimov, line 258: This, btw is a really bad style.

pytest_tests/testsuites/acl/test_eacl.py
```diff
@@ -252,436 +252,6 @@ def test_extended_acl_deny_replication(
...
-                wallet_config=ir_wallet.config_path,
-            )
-            assert can_put_object(
-                storage_wallet.wallet_path,
-                cid,
-                file_path,
-                shell=self.shell,
-                cluster=self.cluster,
-                wallet_config=storage_wallet.config_path,
-            )
```
roman-khimov, line 279: For example, this pair perfectly makes sense as a test for basic ACL, SN should be able to PUT, while IR should not.

pytest_tests/testsuites/acl/test_eacl.py
```diff
@@ -252,436 +252,6 @@ def test_extended_acl_deny_replication(
...
-            assert can_get_object(
-                storage_wallet.wallet_path,
-                cid,
-                object_oids[0],
-                file_path,
-                shell=self.shell,
-                cluster=self.cluster,
-                wallet_config=storage_wallet.config_path,
-            )
```
roman-khimov, line 298: This doesn't make much sense.

pytest_tests/testsuites/acl/test_eacl.py
```diff
@@ -252,436 +252,6 @@ def test_extended_acl_deny_replication(
...
-                wallet_config=ir_wallet.config_path,
-            )
-            assert can_get_head_object(
-                storage_wallet.wallet_path,
-                cid,
-                object_oids[0],
-                shell=self.shell,
-                endpoint=endpoint,
-                wallet_config=storage_wallet.config_path,
-            )
```
roman-khimov, line 315: Like this

pytest_tests/testsuites/acl/test_eacl.py
```diff
@@ -252,436 +252,6 @@ def test_extended_acl_deny_replication(
...
-                wallet_config=ir_wallet.config_path,
-            )
-            assert can_search_object(
-                storage_wallet.wallet_path,
-                cid,
-                shell=self.shell,
-                endpoint=endpoint,
-                oid=object_oids[0],
-                wallet_config=storage_wallet.config_path,
-            )
```
roman-khimov, line 332: And this.

pytest_tests/testsuites/acl/test_eacl.py
```diff
@@ -252,436 +252,6 @@ def test_extended_acl_deny_replication(
...
-                )
-            with pytest.raises(AssertionError):
-                assert can_get_range_of_object(
-                    wallet=storage_wallet.wallet_path,
-                    cid=cid,
-                    oid=object_oids[0],
-                    shell=self.shell,
-                    endpoint=endpoint,
-                    wallet_config=storage_wallet.config_path,
-                )
```
roman-khimov, line 351: And this

pytest_tests/testsuites/acl/test_eacl.py
```diff
@@ -252,436 +252,6 @@ def test_extended_acl_deny_replication(
...
-            )
-
-            assert can_get_range_hash_of_object(
-                wallet=storage_wallet.wallet_path,
-                cid=cid,
-                oid=object_oids[0],
-                shell=self.shell,
-                endpoint=endpoint,
-                wallet_config=storage_wallet.config_path,
-            )
```
roman-khimov, line 369: And this

pytest_tests/testsuites/acl/test_eacl.py
```diff
@@ -252,436 +252,6 @@ def test_extended_acl_deny_replication(
...
-                )
-            with pytest.raises(AssertionError):
-                assert can_delete_object(
-                    wallet=storage_wallet.wallet_path,
-                    cid=cid,
-                    oid=object_oids[0],
-                    shell=self.shell,
-                    endpoint=endpoint,
-                    wallet_config=storage_wallet.config_path,
-                )
```
roman-khimov, line 388: While deletion test has some meaning, none of them should be able to do that.

roman-khimov (changes requested):

roman-khimov:
When will this be updated?
https://http.fs.neo.org/HXSaMJXk2g8C14ht8HSi7BBaiYZ1HeWh2xnWPGQCg4H6/722-1694457866/index.html#suites/69b447004f3c682f67955ec900a09cf0/d419bcbd5433e713/

pytest_tests/testsuites/acl/test_eacl.py
vvarg229: > I don't see anything similar for basic ACL, is that the case?

I don't quite understand this comment. How is basic ACL and these changes related?

vvarg229: > I don't see anything similar for basic ACL, is that the case?

I don't quite understand this comment. How is basic ACL and these changes related?

roman-khimov: They're not directly related, the question is do we have similar tests for basic ACL? If we're not --- let's have an issue for it.

vvarg229: > They're not directly related, the question is do we have similar tests for basic ACL? If we're not --- let's have an issue for it.

https://github.com/nspcc-dev/neofs-testcases/issues/636

roman-khimov: #634. So you want to merge it as is? Frankly, I've expected it to be solved in a day. If that takes too long, OK, let's move on with stripped test and a separate issue.

vvarg229: > #634. So you want to merge it as is? Frankly, I've expected it to be solved in a day. If that takes too long, OK, let's move on with stripped test and a separate issue.

No, I don't want to merge it as is.
But it seems to me that tests for basic acl are a separate task.

I have a question - should basic acl allow changing the system role?
Wouldn't this lead to the same bugs that we fix by removing this functionality from eacl?

And here I will leave the tests that check negative scenarios with system eacl.

roman-khimov: That's the beauty of basic ACL, it was always implemented in a way forbidding messing with system bits. They can only be set to what they need to be set according to the spec.

vvarg229: > > #634. So you want to merge it as is? Frankly, I've expected it to be solved in a day. If that takes too long, OK, let's move on with stripped test and a separate issue.
> 
> No, I don't want to merge it as is. But it seems to me that tests for basic acl are a separate task.
> 
> I have a question - should basic acl allow changing the system role? Wouldn't this lead to the same bugs that we fix by removing this functionality from eacl?
> 
> And here I will leave the tests that check negative scenarios with system eacl.

Fixed

pytest_tests/testsuites/acl/test_eacl.py
vvarg229: > For example, this pair perfectly makes sense as a test for basic ACL, SN should be able to PUT, while IR should not.

Fixed

pytest_tests/testsuites/acl/test_eacl.py
vvarg229: > This doesn't make much sense.

Fixed

pytest_tests/testsuites/acl/test_eacl.py
vvarg229: > Like this

Fixed

pytest_tests/testsuites/acl/test_eacl.py
vvarg229: > And this.

Fixed

pytest_tests/testsuites/acl/test_eacl.py
vvarg229: > And this

Fixed

pytest_tests/testsuites/acl/test_eacl.py
vvarg229: > While deletion test has some meaning, none of them should be able to do that.

Fixed

pytest_tests/testsuites/acl/test_eacl.py
vvarg229: > And this

Fixed

roman-khimov (approved):
A bit too many commits than needed (at least 47be96fb6495c271f1c597de8fd37690d277d70c and 15be7ce9497f83dedc642e2b1c121a08efb51b53 can be squashed), but OK.

PR #628, opened 2023-09-06
tests: add tests for network config change via neofs-adm

evgeniiz321:
The following issues were found:

1. ContainerAliasFee is not shown in netmap netinfo:

```
COMMAND: neofs-cli --config ./TemporaryDir/empty-password.yml netmap netinfo --rpc-endpoint 's01.neofs.devenv:8080' --wallet '../neofs-dev-env/services/storage/wallet01.json'
RETCODE: 0

STDOUT:
Epoch: 5
Network magic: [net 0x3c2d] 15405
Time per block: 1s
NeoFS network configuration (system)
  Audit fee: 10000
  Storage price: 100000000
  Container fee: 1000
  EigenTrust alpha: 0.1
  Number of EigenTrust iterations: 4
  Epoch duration: 240
  Inner Ring candidate fee: 10000000000
  Maximum object size: 524288
  Withdrawal fee: 100000000
  Homomorphic hashing disabled: false
  Maintenance mode allowed: false
NeoFS network configuration (other)
  SystemDNS: 636f6e7461696e6572

STDERR:

Start / End / Elapsed	 01:58:25.658936 / 01:58:25.679838 / 0:00:00.020902
```
I'm not sure if this is by design or not, but still.

2. There is the force param, but I'm not sure what is the use case for it. But nevertheless, if I set some unknown_key=120 with this force flag, I get some strange value afterwards:

```
COMMAND: neofs-cli --config ./TemporaryDir/empty-password.yml netmap netinfo --rpc-endpoint 's01.neofs.devenv:8080' --wallet '../neofs-dev-env/services/storage/wallet01.json'
RETCODE: 0

STDOUT:
Epoch: 5
Network magic: [net 0x3c2d] 15405
Time per block: 1s
NeoFS network configuration (system)
  Audit fee: 10000
  Storage price: 100000000
  Container fee: 1000
  EigenTrust alpha: 0.1
  Number of EigenTrust iterations: 4
  Epoch duration: 240
  Inner Ring candidate fee: 10000000000
  Maximum object size: 524288
  Withdrawal fee: 100000000
  Homomorphic hashing disabled: false
  Maintenance mode allowed: false
NeoFS network configuration (other)
  SystemDNS: 636f6e7461696e6572
  unknown_key: 313230
```
unknown_key: 313230, although I set it to 120. Not sure, how the code should parse unknown values. Perhaps it is better to interpret everything as a string.

3. If I set some known key to invalid value, I got error like this:
```
COMMAND: neofs-adm --config /home/runner/work/neofs-node/neofs-node/neofs-testcases/venv.local-pytest/../../neofs-dev-env/neofs-adm.yml morph set-config --rpc-endpoint 'http://morph-chain.neofs.devenv:30333' --alphabet-wallets '../neofs-dev-env/services/ir' HomomorphicHashingDisabled=0.2
RETCODE: 1

STDOUT:
Error: could not parse HomomorphicHashingDisabled's value '0.2' as bool: strconv.ParseBool: parsing "0.2": invalid syntax
```
The error message seems to me a bit awkward.

Allure - [allure-0.tar.gz](https://github.com/nspcc-dev/neofs-testcases/files/12531241/allure-0.tar.gz)

### DIFF
--- a/pytest_tests/testsuites/network/test_config_changes.py
+++ b/pytest_tests/testsuites/network/test_config_changes.py
@@ -1,0 +1,144 @@
+from typing import Union
+
+import allure
+import pytest
+from cluster_test_base import ClusterTestBase
+from common import NEOFS_ADM_CONFIG_PATH, NEOFS_ADM_EXEC
+from neofs_testlib.cli import NeofsAdm
+from python_keywords.neofs_verbs import get_netmap_netinfo
+
+CONFIG_KEYS_MAPPING = {
+    "MaxObjectSize": "maximum_object_size",
+    "BasicIncomeRate": "storage_price",
+    "AuditFee": "audit_fee",
+    "EpochDuration": "epoch_duration",
+    "ContainerFee": "container_fee",
+    "EigenTrustIterations": "number_of_eigentrust_iterations",
+    "EigenTrustAlpha": "eigentrust_alpha",
+    "InnerRingCandidateFee": "inner_ring_candidate_fee",
+    "WithdrawFee": "withdrawal_fee",
+    "HomomorphicHashingDisabled": "homomorphic_hashing_disabled",
+    "MaintenanceModeAllowed": "maintenance_mode_allowed",
+}
+
+
+@allure.title("Network configuration changes via neofs-adm")
+@pytest.mark.network_config
+class TestNetworkConfigChange(ClusterTestBase):
+    @pytest.mark.parametrize(
+        "key, value",
+        [
+            ("MaxObjectSize", 1048576),
+            ("BasicIncomeRate", 50000000),
+            ("AuditFee", 5000),
+            ("EpochDuration", 480),
+            ("ContainerFee", 2000),
+            pytest.param(
+                "ContainerAliasFee",
+                5000,
+                marks=pytest.mark.skip(
+                    reason="https://github.com/nspcc-dev/neofs-testcases/issues/630"
+                ),
+            ),
+            ("EigenTrustIterations", 8),
+            ("EigenTrustAlpha", 0.2),
+            ("InnerRingCandidateFee", 5000000000),
+            ("WithdrawFee", 200000000),
+            ("HomomorphicHashingDisabled", True),
+            ("MaintenanceModeAllowed", True),
+        ],
+    )
+    def test_config_update_single_value(
+        self, key: str, value: Union[str, int, bool], clean_config: None
+    ):
+        allure.dynamic.title(f"Set '{key}' to '{value}'")
+        self._set_and_verify_config_keys(**{key: value})
+
+    @allure.title("Multiple network config keys can be changed at once")
+    def test_config_update_multiple_values(self, clean_config: None):
+        new_key_value_pairs = {
+            "MaxObjectSize": 1048576,
+            "BasicIncomeRate": 50000000,
+            "AuditFee": 5000,
+        }
+        self._set_and_verify_config_keys(**new_key_value_pairs)
+
+    @pytest.mark.parametrize(
+        "key, value",
+        [
+            ("MaxObjectSize", "VeryBigSize"),
+            ("BasicIncomeRate", False),
+            ("HomomorphicHashingDisabled", 0.2),
+        ],
+    )
+    @allure.title("Set network config key to invalid value")
+    def test_config_set_invalid_value(self, key: str, value: Union[str, int, bool]):
+        self._set_and_verify_config_keys(**{key: value})
+
+    @allure.title("Set multiple network config keys to invalid values with force")
+    def test_config_set_multiple_invalid_values(self):
+        self._set_and_verify_config_keys(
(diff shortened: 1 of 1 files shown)


pytest_tests/testsuites/network/test_config_changes.py
```diff
@@ -0,0 +1,151 @@
...
+from neofs_testlib.cli import NeofsAdm
+from python_keywords.neofs_verbs import get_netmap_netinfo
+
+CONFIG_KEYS_MAPPING = {
+    "MaxObjectSize": "maximum_object_size",
+    "BasicIncomeRate": "storage_price",
+    "AuditFee": "audit_fee",
+    "EpochDuration": "epoch_duration",
+    "ContainerFee": "container_fee",
+    # "ContainerAliasFee": "",
```
vvarg229, line 16: Maybe remove the comment?

pytest_tests/testsuites/network/test_config_changes.py
```diff
@@ -0,0 +1,151 @@
...
+class TestNetworkConfigChange(ClusterTestBase):
+    @pytest.mark.parametrize(
+        "key, value",
+        [
+            ("MaxObjectSize", 1048576),
+            ("BasicIncomeRate", 50000000),
+            ("AuditFee", 5000),
+            ("EpochDuration", 480),
+            ("ContainerFee", 2000),
+            # ("ContainerAliasFee",),
```
vvarg229, line 38: Mb remove the comment here too?

pytest_tests/testsuites/network/test_config_changes.py
```diff
@@ -0,0 +1,151 @@
...
+                wallet_config=storage_node.get_wallet_config_path(),
+                endpoint=storage_node.get_rpc_endpoint(),
+                shell=storage_node.host.get_shell(),
+            )
+            for key in new_key_value_pairs.keys():
+                original_key_value_pairs[key] = net_info[CONFIG_KEYS_MAPPING[key]]
+
+        self._set_and_verify_config_keys(**new_key_value_pairs)
+
+        with allure.step("Change network config back to original state"):
```
vvarg229, line 89: It seems to me that changing the config to the original state is better made as a fixture

vvarg229 (changes requested):

pytest_tests/testsuites/network/test_config_changes.py
evgeniiz321: It is actually philosophical question here, I consider this action as a part of the test, not a teardown. And fixtures are usually something that don't really belong to a test. I can move it to a fixture, but then it will be dissappeared from the test body, and will be a part of teardown.

pytest_tests/testsuites/network/test_config_changes.py
evgeniiz321: I am waiting for an answer to my description questions. I'm not sure if 'ContainerAliasFee' is not shown by design or if it is a bug. I will either uncomment it and skip as a bug or just remove completely, depends on an answer

pytest_tests/testsuites/network/test_config_changes.py
evgeniiz321: The same situation as above

pytest_tests/testsuites/network/test_config_changes.py
roman-khimov: Leave it, but create an issue that depends on https://github.com/nspcc-dev/neofs-node/issues/2553 (which in turn depends on https://github.com/nspcc-dev/neofs-sdk-go/issues/514).

evgeniiz321: what about problems 2 and 3 in the description? are they bugs as well, or I should tolerate this behavior?

pytest_tests/testsuites/network/test_config_changes.py
vvarg229: > It is actually philosophical question here, I consider this action as a part of the test, not a teardown. And fixtures are usually something that don't really belong to a test. I can move it to a fixture, but then it will be dissappeared from the test body, and will be a part of teardown.

If any of the test steps above fail, the network config  will not go back. This may lead to incorrect execution of next tests.

evgeniiz321: Fixed.

pytest_tests/testsuites/network/test_config_changes.py
evgeniiz321: Removed comment, added skipped param

roman-khimov:
As for 2: I'd say it's not a bug, you have to use force for it, so likely you know what you're doing and parameter is passed as is in this case (and displayed in hex, who knows what's inside?) Maybe some documentation can be improved for adm to explain it, file an issue for node.

3: The main thing is that it works as intended (value reject), but the message can be improved, file an issue as well.

roman-khimov (approved):

vvarg229 (approved):

PR #647, opened 2023-09-23
lib: use random values for ranges, closes #632

### LINKED ISSUE #632: GETRANGE and GETRANGEHASH have the same requested length
I'm always frustrated when I need to debug RANGE and RANGEHASH requests. A node may produce its own request to satisfy a user's one, and it may be a different request kind (e.g. user's RANGEHASH may produce RANGE). Logs do not allow understanding of what exact request happened cause both RANGE and RANGEHASH have the same length ([here](https://github.com/nspcc-dev/neofs-testcases/blob/7af282f90f32e80301a146c91d9db49ee2bcb0b8/robot/resources/lib/python_keywords/object_access.py#L162) and [here](https://github.com/nspcc-dev/neofs-testcases/blob/7af282f90f32e80301a146c91d9db49ee2bcb0b8/robot/resources/lib/python_keywords/object_access.py#L193)) to the same object usually.

If it does not break other tests, I would like to ask to make ranges different for the different requests (e.g. `[0:9]` and `[0:10]` or even a rand length for a different test suite in the whole test run). They are often run one by one and network communication becomes a mess (a lot of [0:10] RANGE requests).

## Describe alternatives you've considered
Do nothing. I will change those parameters manually every time I run RANGE and RANGEHASH tests then.

## Additional context
Sounds strange but would have saved some time in https://github.com/nspcc-dev/neofs-node/issues/2541 debug.

evgeniiz321:


vvarg229 (approved):

robot/resources/lib/python_keywords/object_access.py
```diff
@@ -142,6 +143,12 @@ def can_get_head_object(
     return True
 
 
+def _generate_random_range_cut(start: int = 0, end: int = 10):
+    start = random.randint(start, end)
+    end = random.randint(start, end)
+    return f"{start}:{end}"
```
roman-khimov, line 149: What if it's 10:10 or 0:0?

evgeniiz321: I assumed these are valid ranges, at least the command works:

```
COMMAND: neofs-cli --config /home/runner/work/neofs-node/neofs-node/neofs-testcases/wallet_config.yml object range --rpc-endpoint 's01.neofs.devenv:8080' --wallet '/home/runner/work/neofs-node/neofs-node/neofs-testcases/TemporaryDir/9565822e-6a8a-4dd6-8191-d1e79649b0bd.json' --cid '61M7pv5kkku8yNhrLiohmLd31XuSDCR3WLQkiGrd4Sgd' --oid 'B5vP6r1BQ3ktxY2wfvDNwBy6UnKYTRAQRCCqcCzUgGsb' --range '10:10' --file 'TemporaryDir/TestObjectsDir/3bbbe60d-6fdd-4b04-9449-0fd2e8826521'
RETCODE: 0

STDOUT:
[TemporaryDir/TestObjectsDir/3bbbe60d-6fdd-4b04-9449-0fd2e8826521] Payload successfully saved

STDERR:

Start / End / Elapsed	 19:31:43.302994 / 19:31:43.707277 / 0:00:00.404283
```

carpawell, line 149: that is not `[start:end]` meaning but is `[offset:height]` in fact. so i would like to ask you to rename it accordingly to prevent any misunderstanding. also, we do not allow `[X:0]` requests do we have any tests for it? just to fix it

btw:
```
▶ neofs-cli object range -h | grep '\-\-range'
      --range string          Range to take data from in the form offset:length
```

so we are not that bad and try to doc it (i can imagine we would have written it as just "Range to take data from"). but i agree that slicing format (`[start:end]`) may be more presentative

carpawell (changes requested):

robot/resources/lib/python_keywords/object_access.py
evgeniiz321: for additional tests - https://github.com/nspcc-dev/neofs-testcases/issues/652
fixed naming and the code accordingly

evgeniiz321: fixed according to Pavel's comment

robot/resources/lib/python_keywords/object_access.py
```diff
@@ -142,6 +143,13 @@ def can_get_head_object(
     return True
 
 
+def _generate_random_range_cut(offset: int = 0, height: int = 10):
```
roman-khimov, line 146: `s/height/length/`?

roman-khimov (commented):
OK otherwise.

robot/resources/lib/python_keywords/object_access.py
evgeniiz321: renamed.

roman-khimov (approved):

PR #655, opened 2023-09-26
acl: Test basic ACL IR and STORAGE rules compliance

vvarg229:
Added a test to verify compliance rules with the basic ACL IR and STORAGE. This is a copy of a similar test from eACL.

Closes https://github.com/nspcc-dev/neofs-testcases/issues/634

### DIFF
--- a/pytest_tests/testsuites/acl/test_acl.py
+++ b/pytest_tests/testsuites/acl/test_acl.py
@@ -1,6 +1,10 @@
 import allure
 import pytest
+
+from testsuites.acl.conftest import Wallets
 from cluster_test_base import ClusterTestBase
+from object_access import can_put_object, can_get_object, can_get_head_object, can_search_object, \
+    can_get_range_of_object, can_get_range_hash_of_object, can_delete_object
 from python_keywords.acl import EACLRole
 from python_keywords.container import create_container
 from python_keywords.container_access import (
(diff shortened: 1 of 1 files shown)


pytest_tests/testsuites/acl/test_acl.py
```diff
@@ -178,3 +180,149 @@ def test_basic_acl_readonly(self, wallets, client_shell, read_only_container, fi
...
+
+        cid = public_container
+
+        with allure.step("Add test objects to container"):
+            owner_object_oid = put_object_to_random_node(
+                user_wallet.wallet_path, file_path, cid, shell=self.shell, cluster=self.cluster
+            )
+
+        with allure.step("Check IR and STORAGE rules compliance"):
+            assert not can_put_object(
```
evgeniiz321, line 203: Can we use ```check_full_access_to_container (robot/resources/lib/python_keywords/container_access.py)``` or ```check_no_access_to_container``` here? or some modification of it maybe?

pytest_tests/testsuites/acl/test_acl.py
```diff
@@ -178,3 +180,149 @@ def test_basic_acl_readonly(self, wallets, client_shell, read_only_container, fi
...
+                storage_wallet.wallet_path,
+                cid,
+                shell=self.shell,
+                endpoint=endpoint,
+                oid=owner_object_oid,
+                wallet_config=storage_wallet.config_path,
+            )
+
+            with pytest.raises(AssertionError):
+                assert can_get_range_of_object(
```
evgeniiz321, line 274: AssertionError is expected to be raised from the assert here? Or from somewhere internally? If from this assert, maybe better to use just ```assert no can_get...```

evgeniiz321: Also it is not clear from the test why some operations are allowed and others not. It is better to add some related comment

vvarg229: > AssertionError is expected to be raised from the assert here? Or from somewhere internally? If from this assert, maybe better to use just `assert not can_get...`

Fixed

vvarg229: > Also it is not clear from the test why some operations are allowed and others not. It is better to add some related comment

Fixed

pytest_tests/testsuites/acl/test_acl.py
evgeniiz321: They've already been put there. ```check_full_access_to_container``` looks pretty much the same as these conditions here with some variations. But ok, we can leave it as is

pytest_tests/testsuites/acl/test_acl.py
```diff
@@ -178,3 +180,180 @@ def test_basic_acl_readonly(self, wallets, client_shell, read_only_container, fi
                 shell=client_shell,
                 cluster=self.cluster,
             )
+
+    @allure.title("Test basic ACL IR and STORAGE rules compliance")
+    def test_basic_acl_ir_storage_rules_compliance(self, wallets, public_container, file_path):
```
evgeniiz321, line 185: No type annotations here, at least some can be added

pytest_tests/testsuites/acl/test_acl.py
vvarg229: > looks pretty much the same

But it's not full access here

pytest_tests/testsuites/acl/test_acl.py
vvarg229: > No type annotations here, at least some can be added

Thank you very much! Sorry, I wasn't paying attention.
Fixed.

evgeniiz321 (approved):

PR #657, opened 2023-09-28
tests: add test for special http headers, closes #651

### LINKED ISSUE #651: Special header tests for HTTP
x-object, x-container, x-owner. Probably some old tests can be enhanced to check for them.
https://github.com/nspcc-dev/neofs-http-gw#replies

evgeniiz321:
X-Owner-Id is always equal to the address of the wallet specified during the http gw deployment, in case of our dev-env - https://github.com/nspcc-dev/neofs-dev-env/blob/master/services/http_gate/wallet.json#L5. Is it expected? As for me, I would expect X-Owner-Id to be equal to an address of a wallet used during a container creation.

### DIFF
--- a/pytest_tests/testsuites/services/http_gate/test_http_gate.py
+++ b/pytest_tests/testsuites/services/http_gate/test_http_gate.py
@@ -1,8 +1,10 @@
+import json
 import logging
 import os
 
 import allure
 import pytest
+from cluster import Cluster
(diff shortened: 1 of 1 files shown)


roman-khimov:
> Is it expected?

Yes, since, objects are created by the gateway and object owner can differ from the container owner.

roman-khimov (approved):

PR #658, opened 2023-09-28
tests: verify COMMON_PREFIX search filter, closes #244

### LINKED ISSUE #244: `COMMON_PREFIX` matchtype coverage
https://github.com/nspcc-dev/neofs-api/issues/164
https://neospcc.atlassian.net/wiki/spaces/NEOF/pages/2132410369/COMMON+PREFIX+attribute+matchtype+v0.23.1+v0.24.0

evgeniiz321:


pytest_tests/testsuites/object/test_object_api.py
```diff
@@ -337,6 +341,69 @@ def test_object_search_should_return_tombstone_items(
...
+            with allure.step(f"Search objects by path: {common_prefix}"):
+                result = cli.object.search(
+                    rpc_endpoint=self.cluster.default_rpc_endpoint,
+                    wallet=wallet,
+                    cid=cid,
+                    filters=[f"{FILEPATH_ATTR_NAME} COMMON_PREFIX {common_prefix}"],
+                    root=True,
+                )
+
+                found_oids = re.findall(r"(\w{43,44})", result.stdout)
```
roman-khimov, line 381: This pattern should be somewhere in the lib probably. Here particularly it's also a whole line.

pytest_tests/testsuites/object/test_object_api.py
```diff
@@ -337,6 +341,69 @@ def test_object_search_should_return_tombstone_items(
...
+                if len(found_oids) > 1:
+                    assert expected_oids == sorted(found_oids)
+                elif len(found_oids):
+                    assert expected_oids == found_oids[0]
+                else:
+                    raise AssertionError(f"No objects found by common prefix: {common_prefix}")
+        
+        for common_prefix in (
```
vvarg229, line 389: Maybe this would be better organized as an enumeration?

pytest_tests/testsuites/object/test_object_api.py
```diff
@@ -337,6 +341,69 @@ def test_object_search_should_return_tombstone_items(
...
 
+    @allure.title("Validate objects search by common prefix")
+    def test_search_object_api_common_prefix(self, default_wallet: str, simple_object_size: int):
+        FILEPATH_ATTR_NAME = "FilePath"
+        cli = NeofsCli(self.shell, NEOFS_CLI_EXEC, WALLET_CONFIG)
+        wallet = default_wallet
+        cid = create_container(wallet, shell=self.shell, endpoint=self.cluster.default_rpc_endpoint)
+
+        objects = {}
+        for _ in range(5):
```
vvarg229, line 352: Looks like a magic number

pytest_tests/testsuites/object/test_object_api.py
```diff
@@ -337,6 +341,69 @@ def test_object_search_should_return_tombstone_items(
                     object_type == "TOMBSTONE"
                 ), f"Object wasn't deleted properly. Found object {tombstone_oid} with type {object_type}"
 
+    @allure.title("Validate objects search by common prefix")
+    def test_search_object_api_common_prefix(self, default_wallet: str, simple_object_size: int):
+        FILEPATH_ATTR_NAME = "FilePath"
```
vvarg229, line 346: Maybe it would be better to move it to common?

pytest_tests/testsuites/object/test_object_api.py
```diff
@@ -337,6 +341,69 @@ def test_object_search_should_return_tombstone_items(
...
+        cli = NeofsCli(self.shell, NEOFS_CLI_EXEC, WALLET_CONFIG)
+        wallet = default_wallet
+        cid = create_container(wallet, shell=self.shell, endpoint=self.cluster.default_rpc_endpoint)
```
vvarg229, line 349: Wouldn't it be better to do container creation in fixture?

pytest_tests/testsuites/object/test_object_api.py
```diff
@@ -337,6 +341,69 @@ def test_object_search_should_return_tombstone_items(
...
+
+        objects = {}
+        for _ in range(5):
+            file_path = generate_file(simple_object_size)
+
+            with allure.step("Put objects"):
+                objects[file_path] = put_object_to_random_node(
```
vvarg229, line 356: The put_object_to_random_node function returns the ID.
Why do you store the ID in objects[file_path]?

vvarg229 (changes requested):

pytest_tests/testsuites/object/test_object_api.py
evgeniiz321: no, it wouldn't =) it is just test cases inside this particular test, it can be used inside a parametrization decorator, but I decided to move it here to speed up test a little bit

pytest_tests/testsuites/object/test_object_api.py
evgeniiz321: It is definitely a magic number that is used inside this test in one place. I can replace it with a string but it would be just one more line without a reason right now.

pytest_tests/testsuites/object/test_object_api.py
evgeniiz321: I don't think so - it is not some 'special' attribute, you can use abrakadabra as well here

pytest_tests/testsuites/object/test_object_api.py
evgeniiz321: Further in test I'm using this mapping to get an ID by file_path to verify an output from the search method

pytest_tests/testsuites/object/test_object_api.py
evgeniiz321: Perhaps, but for some reason containers are not deleted in this test suite (maybe it consumes time and is not really needed cause doesn't affect other tests), so the fixture will be useless, just with one statement for creation and without a proper cleanup.

pytest_tests/testsuites/object/test_object_api.py
evgeniiz321: fixed

pytest_tests/testsuites/object/test_object_api.py
vvarg229: > Perhaps, but for some reason containers are not deleted in this test suite (maybe it consumes time and is not really needed cause doesn't affect other tests), so the fixture will be useless, just with one statement for creation and without a proper cleanup.

Maybe we should start deleting containers in this test suite?
Not deleting containers and other test files in tests is many places in the code, it's a very bad practice, let's make the test code better.

pytest_tests/testsuites/object/test_object_api.py
vvarg229: > It is definitely a magic number that is used inside this test in one place. I can replace it with a string but it would be just one more line without a reason right now.

For debugging in the future, this would be a good reference

evgeniiz321: fixed

pytest_tests/testsuites/object/test_object_api.py
evgeniiz321: fixed for this test

vvarg229 (approved):

PR #669, opened 2023-11-01
helpers, s3_gate: Fix acl tests

vvarg229:
The neofs-s3-gw repository has made changes to how acl works. This commit changes the tests so that they work correctly with the new acl rules.

Tests result:
https://http.fs.neo.org/7ZVSQHtioVRiDT2btN9xEmQhczXV9vryCEAHUawfD8Sc/126-1700220863/index.html

Important:
Tests on the current master will crash, this change will only work with a recent s3-gw

### DIFF
--- a/pytest_tests/helpers/s3_helper.py
+++ b/pytest_tests/helpers/s3_helper.py
@@ -12,6 +12,45 @@ from dateutil.parser import parse
 logger = logging.getLogger("NeoLogger")
 
 
+ACL_TO_PERMISSION_MAP_BUCKET = {
+    "Group": {
+        "public-read-write": "FULL_CONTROL",
+        "grant-full-control": "FULL_CONTROL",
+        "public-read": "READ",
+        "grant-read": "READ",
+        "authenticated-read": "READ",
+        "grant-write": "WRITE",
+    },
+    "CanonicalUser": {
+        "public-read-write": "FULL_CONTROL",
+        "grant-full-control": "FULL_CONTROL",
+        "public-read": "FULL_CONTROL",
+        "grant-read": "FULL_CONTROL",
+        "authenticated-read": "FULL_CONTROL",
+        "grant-write": "FULL_CONTROL",
+        "private": "FULL_CONTROL",
+    },
+}
+
+ACL_TO_PERMISSION_MAP_OBJECT = {
+    "Group": {
+        "public-read-write": "FULL_CONTROL",
+        "grant-full-control": "FULL_CONTROL",
+        "public-read": "READ",
+        "grant-read": "READ",
+        "authenticated-read": "READ",
+    },
+    "CanonicalUser": {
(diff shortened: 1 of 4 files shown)


pytest_tests/helpers/s3_helper.py
```diff
@@ -12,6 +12,37 @@
...
+    },
+}
+
+ACL_TO_PERMISSION_MAP_OBJECT = {
+    "Group": {"public-read-write": "WRITE", "public-read": "READ", "grant-read": "READ"},
+    "CanonicalUser": {
+        "public-read-write": "FULL_CONTROL",
+        "public-read": "WRITE",
+        "grant-read": "WRITE",
+        "private": "WRITE",
```
smallhive, line 41: AWS documentation is so messy, but according to [Canned ACL](https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html), I tend to think, these rows should be FULL_CONTROL as well

vvarg229: > AWS documentation is so messy, but according to [Canned ACL](https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html), I tend to think, these rows should be FULL_CONTROL as well

Then you need to fix that in the s3-gw repo too

smallhive: That's the way. But I think we should discuss it with @evgeniiz321, to be sure we are on the same page with documentation understanding.

vvarg229: > AWS documentation is so messy, but according to [Canned ACL](https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html), I tend to think, these rows should be FULL_CONTROL as well

Fixed

roman-khimov:
Rebase, fix problems and let's move on.

pytest_tests/testsuites/services/s3_gate/test_s3_object.py
```diff
@@ -734,7 +737,9 @@ def test_s3_put_object_acl(
                 self.s3_client, bucket, file_path_4, ACL="authenticated-read"
             )
             obj_acl = s3_gate_object.get_object_acl_s3(self.s3_client, bucket, file_name)
-            assert_s3_acl(acl_grants=obj_acl, permitted_users="AllUsers")
+            assert_object_s3_acl(
+                acl_grants=obj_acl, permitted_users="AllUsers", acl="authenticated-read"
```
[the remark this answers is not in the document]
vvarg229: > DRY. You have these ACL names already.

Fixed

vvarg229:
> Rebase, fix problems and let's move on.

Fixed, tests result: https://http.fs.neo.org/7ZVSQHtioVRiDT2btN9xEmQhczXV9vryCEAHUawfD8Sc/126-1700220863/index.html

evgeniiz321 (approved):

PR #680, opened 2023-12-18
Update some s3 tests to use dynamic env

evgeniiz321:
This is an example of how we can move tests to use dynamic environment instead of a static dev-env. In parallel we can refactor existing non optimal/legacy code.

1. Separate directory for new tests and library code - dynamic_env_pytest_tests
2. In the meantime we reuse some existing code located in pytest_tests/* robot/*
3. During tests moving to a new dynamic env - we refactor existing legacy code and move it to  dynamic_env_pytest_tests/lib as well.
4. For new tests at dynamic_env_pytest_tests we create a separate validation inside this repo


vvarg229 (changes requested):
I think it's better to separate into different commits the CI file, the "library" for the dynamic_env_pytest_tests and new tests.

robot/resources/files/container_policy.json
```diff
@@ -0,0 +1,3 @@
+{
+    "rep-1": "REP 1"
+}
```
vvarg229, line 3: New line?

requirements.txt
```diff
@@ -52,7 +52,7 @@ requests==2.31.0
 robotframework==4.1.2
 s3transfer==0.3.7
 six==1.16.0
-tenacity==8.0.1
+tenacity>=8.0.1
```
vvarg229, line 55: It's a good idea to write why in a separate commit message.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.isort]
 profile = "black"
-src_paths = ["pytest_tests", "robot"]
+src_paths = ["dynamic_env_pytest_tests", "pytest_tests", "robot"]
```
vvarg229, line 3: By the way, maybe delete "robot"?

vvarg229 (changes requested):

robot/resources/files/container_policy.json
```diff
@@ -0,0 +1,3 @@
+{
+    "rep-1": "REP 1"
```
vvarg229, line 2: Maybe add the number of REP greater than 1?

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,246 @@
...
+permissions: write-all
+
+jobs:
+  run_system_tests:
+    runs-on: ubuntu-latest
+    timeout-minutes: 500
+    steps:
+      - name: Get the current date
+        id: date
+        run: echo "::set-output name=timestamp::$(date +%s)"
```
vvarg229, line 33: echo "timestamp=$(date +%s)" >> $GITHUB_OUTPUT

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,246 @@
...
+
+      - name: Set RUN_ID
+        env:
+          TIMESTAMP: ${{ steps.date.outputs.timestamp }}
+        run: echo "RUN_ID=${{ github.run_number }}-$TIMESTAMP" >> $GITHUB_ENV
+
+      - name: Checkout neofs-testcases repository
+        uses: actions/checkout@v3
```
vvarg229, line 41: v4

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,246 @@
...
+        with:
+          path: neofs-testcases
+
+      - name: Checkout neofs-node repository
+        uses: actions/checkout@v3
+        with:
+          repository: nspcc-dev/neofs-node
+          ref: 'master'
```
vvarg229, line 49: Mb latest release?

vvarg229 (changes requested):

pyproject.toml
evgeniiz321: current tests with old dev-env will be broken then

robot/resources/files/container_policy.json
evgeniiz321: it should be added when it is needed, for current pr, there is no need to add more than 1

.github/workflows/run-tests.yml
evgeniiz321: It is a good question. On the one side - 'master' will allow us to always keep tests in 'current' state, since we won't be able to merge if smth is broken. On the other side - fixed release version will allow us to prevent unnecessary building here + tests will be more stable and consistent, since we run them on a well-known version.
@roman-khimov what do you think?

robot/resources/files/container_policy.json
evgeniiz321: fixed

requirements.txt
evgeniiz321: fixed

.github/workflows/run-tests.yml
evgeniiz321: fixed

evgeniiz321:
commits were separated

vvarg229 (approved):

.github/workflows/run-tests.yml
evgeniiz321: I added fixed release versions for every binary except s3 gw, current available release is too old, so I used fixed commit.

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,246 @@
...
+# activate = 1
+#
+# [legacy_sect]
+# activate = 1
+      - name: Fix OpenSSL ripemd160
+        run: |
+          sudo python ./tools/src/openssl_config_fix.py
+        working-directory: neofs-testcases
+
+      - name: Build neofs-node
```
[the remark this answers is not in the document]
evgeniiz321: Fixed, now artifacts are downloaded instead of being built.

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,250 @@
...
+# activate = 1
+#
+# [legacy_sect]
+# activate = 1
+      - name: Fix OpenSSL ripemd160
+        run: |
+          sudo python ./tools/src/openssl_config_fix.py
+        working-directory: neofs-testcases
+
+      - name: Build neofs-s3-gw
```
vvarg229, line 165: Why not make a binary download here?

vvarg229 (changes requested):

.github/workflows/run-tests.yml
evgeniiz321: I need a specific commit here, not a release version

vvarg229 (approved):

PR #719, opened 2024-01-31
dyn_env: adapt object tests

evgeniiz321:


evgeniiz321:
@roman-khimov question regarding this test scenario:

1. create an object
2. lock this object
3. get list of nodes with this object (e.g. we got N nodes)
4. stop storage nodes, remove metabase file
5. start storage nodes again
6. get list of nodes with the object - now we get either N-1 or N+1 nodes

Test assumes that the list of nodes with this object before/after metabase removal should be the same. Is it really the case? Or the object can somehow migrate from a one node to another after nodes stop/start due to some other internal processes?

roman-khimov:
> Or the object can somehow migrate from a one node to another after nodes stop/start due to some other internal processes?

If they all start/stop at the same time then no, there won't be any migration. At the same time technically they never start/stop ideally simultaneously, so there is a chance for some migration during the process (especially if we're talking about GH runners that are running on busy machines). This migration should _increase_ the number of copies though, if you have four SNs and three replicas then they're placed optimally during the initial PUT (say, 1-2-3). Now this set of nodes will not delete the object, but a node can see that some neighbor is out and try replicating onto the remaining one (4).

It can be mitigated by some firewall rule with immediate network connectivity failure/restore for everyone. This will make the test run longer though.

PR #720, opened 2024-01-31
Acl tests new env

vvarg229:


evgeniiz321:
transferred to https://github.com/nspcc-dev/neofs-testcases/pull/721

PR #729, opened 2024-02-21
dyn_env: adapt failovers and load tests

evgeniiz321:
Failover and load tests need some clarification on their purpose and what/how we want to deal with them.

1. Most of the failover tests require a setup with different hosts (physical servers/VMs) and check something strange with strange approaches. There are currently these scenarios:

-  "Block storage nodes traffic using iptables". As I understood we want to check replication logic there. It doesn't really work in its current version and should be rewritten. What I suggest - stop one of the storage node processes and ensure the object is replicated.
- "RPC reconnection". It is very strange to me as well. But as I understood - it wants to check the case when one of the inner ring nodes goes down and that storage nodes start using another inner ring node that is alive.
- "Storage nodes panic" - I'm not sure what is the difference from the case where we just stop the storage node process. 

To sum up, I would suggest to have these new failover tests instead of old ones:
1. Kill storage node processes - ensure objects are replicated. Check different replication policies
2. Multiple inner ring nodes - try to kill one and see that we still can put/read objects
3. tbd
---
2. Load test seems to be very irrelevant to current tests. And current test does not really check anything. We just run k6 and check RC, no results are parsed. As I understood @MaxGelbakhiani runs load tests somewhere else and doesn't rely on this particular test. I would suggest to remove it from current tests, since it doesn't really check anything.  Or we need to figure out what we want from this test. Maybe we can reimplement it and use like a benchmark to see a performance degradation on the pull requests and stop such pull requests where we see such a degradation.

### DIFF
--- a/dynamic_env_pytest_tests/lib/helpers/node_management.py
+++ b/dynamic_env_pytest_tests/lib/helpers/node_management.py
@@ -239,6 +239,29 @@ def check_node_in_map(
     ), f"Expected node with key {node_netmap_key} to be in network map"
 
 
+@allure.step("Wait for storage nodes returned to cluster")
+def wait_all_storage_nodes_returned(neofs_env: NeoFSEnv) -> None:
+    sleep_interval, attempts = 15, 20
+    for _ in range(attempts):
+        if is_all_storage_nodes_returned(neofs_env):
+            return
+        time.sleep(sleep_interval)
+    raise AssertionError("Storage node(s) is broken")
+
+
+def is_all_storage_nodes_returned(neofs_env: NeoFSEnv) -> bool:
+    with allure.step("Run health check for all storage nodes"):
+        for node in neofs_env.storage_nodes:
+            try:
+                health_check = storage_node_healthcheck(node)
+            except Exception as err:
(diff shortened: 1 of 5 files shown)


MaxGelbakhiani:
Yes, my performance test runs don't rely on the current test suite. The approach with running performance and comparing results with the previous realize seems reasonable.

roman-khimov:
Load test was added just a smoke check, the system should survive this test without any issues.

RPC reconnection is more about node Neo RPC, there are lots of problems that can happen when SN loses RPC connection. Now that RPC is tied to IR with embedded CN we can either block/drop this connection using firewall or stop  IR node.

Replication test can be done both by blocking traffic or killing node, from what I see we just do both now and it's good in fact.

`panic_reboot_host` is not bad, but yes, it really needs a VM for that. Can be skipped, `kill -9` is good enough for now.

evgeniiz321:
i've opened two issues to rewrite current load and failovers tests:
https://github.com/nspcc-dev/neofs-testcases/issues/730
https://github.com/nspcc-dev/neofs-testcases/issues/731
this PR will merge in order to get rid of the old dev env code

PR #732, opened 2024-02-21
Fix test for the new wildcard session tokens behavior.

### LINKED ISSUE #728: Container sessions updates
There were some changes in the node (https://github.com/nspcc-dev/neofs-node/pull/2727) that changed the behavior (it was already mentioned in https://github.com/nspcc-dev/neofs-testcases/issues/718). However, there also was some bug about token's `wildcard` field that was fixed in https://github.com/nspcc-dev/neofs-node/pull/2741.

## Current Behavior
https://github.com/nspcc-dev/neofs-testcases/pull/726 changed the error message to expect to the "wrong container" version. After the fix, it should not be possible to get such a message if `wildcard` field is provided (the field means you may not specify a container at all, so "wrong container" does not relate to the situation).

## Expected Behavior
If tests do correct things, no error is expected. I have not seen all the tests but every error case should have some readable message about the error. _No_ "wrong container" for the `wildcard` tokens caused by the bug fixed in https://github.com/nspcc-dev/neofs-node/pull/2741 and also _no_ "timeout" error as it was before since the SN now validates requests before it sends them to the IR. Also, there is exactly one straight happy path (correct container, correct token signer, and so on) and i think any different token usages should be also tested. If anything that can be tested is already tested, it is great then. But if there are some untested cases, i would like them to be tested.

## Possible Solution
1. Run some tests with 0.40.1+ version, see the message. If it is correct (it should describe the "wrong" things the tests sent), adapt the test's expected message. If it is incorrect, create an issue about it and we will fix it.
2. See if anything else can be tested but has not been tested yet.

## Steps to Reproduce
0.40.1. I believe that is all you need.

## Context
If any message is not fully clear, we may adapt it, just create an issue. Also, be ready that some wrong cases _may_ not be reproducible cause tests use `neofs-cli` that may fix some token issues inside.

## Regression
https://github.com/nspcc-dev/neofs-testcases/pull/726. Not sure it had to be merged. It does not adapt the tests, it just fixed (i mean the bug in the node is now "tested and expected") the bug we found.

MaxGelbakhiani:
Fixes #728

Test run workflow and test report with the fixed test may be observed in my fork:
https://github.com/MaxGelbakhiani/neofs-node/actions/runs/7994175346/job/21831643871
https://http.t5.fs.neo.org/9GX3Bvpb5RFapar5g11zwwoBGDdxcSs35764xchZTVUF/26-1708543467/index.html

### DIFF
--- a/pytest_tests/helpers/grpc_responses.py
+++ b/pytest_tests/helpers/grpc_responses.py
@@ -7,6 +7,7 @@ CONTAINER_NOT_FOUND = "code = 3072.*message = container not found"
 # Regex patterns of status codes of Object service (https://github.com/nspcc-dev/neofs-spec/blob/98b154848116223e486ce8b43eaa35fec08b4a99/20-api-v2/object.md)
 MALFORMED_REQUEST = "code = 1024.*message = malformed request"
 WRONG_CONTAINER = "code = 1024.*message = wrong container"
+SESSION_NOT_ISSUED_BY_OWNER = "code = 1024.*message = session was not issued by the container owner"
 OBJECT_ACCESS_DENIED = "code = 2048.*message = access to object operation denied"
 OBJECT_NOT_FOUND = "code = 2049.*message = object not found"
 OBJECT_ALREADY_REMOVED = "code = 2052.*message = object already removed"
--- a/pytest_tests/testsuites/session_token/test_static_session_token_container.py
+++ b/pytest_tests/testsuites/session_token/test_static_session_token_container.py
@@ -1,7 +1,7 @@
 import allure
 import pytest
 from file_helper import generate_file
-from grpc_responses import NOT_SESSION_CONTAINER_OWNER, CONTAINER_DELETION_TIMED_OUT, WRONG_CONTAINER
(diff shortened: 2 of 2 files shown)


roman-khimov (approved):

pytest_tests/helpers/grpc_responses.py
```diff
@@ -7,6 +7,7 @@
 # Regex patterns of status codes of Object service (https://github.com/nspcc-dev/neofs-spec/blob/98b154848116223e486ce8b43eaa35fec08b4a99/20-api-v2/object.md)
 MALFORMED_REQUEST = "code = 1024.*message = malformed request"
 WRONG_CONTAINER = "code = 1024.*message = wrong container"
```
cthulhu-rider, line 9: now unused?

roman-khimov: I'd leave it, it can be useful for new tests.

cthulhu-rider: dont mind

PR #735, opened 2024-03-05
Add tests for the REST gateway

### LINKED ISSUE #723: Test REST gw with HTTP testcases
I'm always frustrated when we don't test our software. REST gateway is not covered by these tests at all right now.

We can easily reuse the HTTP set of tests (except for zip ones) for the REST gw. It's similar in configuration and tests can be reused as is (same API provided). The only caveat is that the REST gw is not currently supported as a service.

tatiana-nspcc:
Close #723

### DIFF
--- a/pytest_tests/tests/services/http_rest_gates/test_http_rest_gate.py
+++ b/pytest_tests/tests/services/http_rest_gates/test_http_rest_gate.py
@@ -34,14 +34,14 @@ OBJECT_NOT_FOUND_ERROR = "not found"
 @allure.link("https://github.com/nspcc-dev/neofs-http-gw#downloading", name="downloading")
 @pytest.mark.sanity
 @pytest.mark.http_and_rest_gates
-class TestHttpGate(NeofsEnvTestBase):
+class TestHttpRestGate(NeofsEnvTestBase):
     PLACEMENT_RULE_1 = "REP 1 IN X CBF 1 SELECT 1 FROM * AS X"
     PLACEMENT_RULE_2 = "REP 2 IN X CBF 2 SELECT 2 FROM * AS X"
 
     @pytest.fixture(scope="class", autouse=True)
     @allure.title("[Class/Autouse]: Prepare wallet and deposit")
     def prepare_wallet(self, default_wallet):
-        TestHttpGate.wallet = default_wallet
+        TestHttpRestGate.wallet = default_wallet
 
     @pytest.fixture(scope="class", params=["HTTP", "REST"])
     def gw_params(self, request):
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -148,7 +148,7 @@ jobs:
         uses: dsaltares/fetch-gh-release-asset@1.1.1
         with:
           repo: 'nspcc-dev/neofs-rest-gw'
-          version: 'tags/v0.6.0'
+          version: 'tags/v0.7.2'
           file: 'neofs-rest-gw-linux-amd64'
           target: 'neofs-testcases/neofs-rest-gw'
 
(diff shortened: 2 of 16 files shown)


tatiana-nspcc:
@evgeniiz321 I've pushed changes for only one file. Could you, please, take a look,  and check if it's the right way?

roman-khimov (commented):
Testlib 1.1.20 doesn't work.

What's the test status for it? Likely zip tests need to be disabled.

requirements.txt
```diff
@@ -29,7 +29,7 @@ mmh3==3.0.0
 multidict==6.0.2
 mypy==0.950
 mypy-extensions==0.4.3
-neofs-testlib==1.1.19
+neofs-testlib==1.1.20
```
evgeniiz321, line 32: it should be 1.1.21

dynamic_env_pytest_tests/tests/services/http_gate/test_http_bearer.py
```diff
@@ -83,15 +83,23 @@ def bearer_token_no_limit_for_others(self, user_container: str) -> str:
             )
             return bearer_token_base64_from_file(bearer_signed)
 
+    @pytest.fixture(scope="class", params=["HTTP", "REST"])
+    def gw_endpoint(self, request):
```
evgeniiz321, line 87: Does it work this way? If it does - that's good. If not, another way is this:
Define a special pytest hook method (put it into conftest.py)
```
def pytest_generate_tests(metafunc):
    if "gw_endpoint" in metafunc.fixturenames:
        metafunc.parametrize("gw_endpoint", ["HTTP, "REST"], indirect=True)
```
And then this fixture will be the same, but without params and you can also define it in conftest.py

requirements.txt
tatiana-nspcc: Fixed.

dynamic_env_pytest_tests/tests/services/http_gate/test_http_headers.py
```diff
@@ -67,7 +77,7 @@ def storage_objects_with_attributes(self, request: FixtureRequest) -> list[Stora
             storage_object_id = upload_via_http_gate_curl(
                 cid=cid,
                 filepath=file_path,
-                endpoint=f"http://{self.neofs_env.http_gw.address}",
+                endpoint=f"http://{self.neofs_env.http_gw.address}",  # ???,
```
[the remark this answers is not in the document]
evgeniiz321: When you tried to interrupt those frozen tests, what was the traceback? Did it fail on REST or HTTP? From what I see, it can be a bug with the rest gateway.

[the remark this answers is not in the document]
tatiana-nspcc: Yep, you're right it relates to bad request response due to "@" symbols in attributes.

pytest_tests/tests/services/http_rest_gates/test_http_rest_gate.py
```diff
@@ -33,15 +33,29 @@
...
     PLACEMENT_RULE_2 = "REP 2 IN X CBF 2 SELECT 2 FROM * AS X"
 
     @pytest.fixture(scope="class", autouse=True)
     @allure.title("[Class/Autouse]: Prepare wallet and deposit")
     def prepare_wallet(self, default_wallet):
-        TestHttpGate.wallet = default_wallet
+        TestHttpRestGate.wallet = default_wallet
+
+    @pytest.fixture(scope="class", params=["HTTP", "REST"])
+    def gw_params(self, request):
```
evgeniiz321, line 47: you can move this fixture to the conftest.py and put it here - tests/services/http_rest_gates/conftest.py, so you don't need to write same fixture in multiple modules

tatiana-nspcc: Done.

PR #756, opened 2024-03-25
tests: new tests for numeric eacl rules

evgeniiz321:


roman-khimov (commented):
Can we also have some special eacls that touch system attributes like epoch, expiration and object size? We also want to check PUTs. Like: you can put objects of size less than X (2*big object size and try to put small, big and 3*big) and expiration less than Y.

evgeniiz321:
@roman-khimov added missing scenarios.

pytest_tests/tests/acl/test_eacl_filters.py
```diff
@@ -627,3 +657,556 @@ def test_extended_acl_allow_filters_object(
...
+
+        with allure.step(f"GET object with numeric value attribute should be allowed"):
+            assert can_get_object(
+                user_wallet.wallet_path,
+                cid,
+                oid,
+                file_path,
+                self.shell,
+                neofs_env=self.neofs_env,
+            ), f"GET is not allowed for this object, while it shouldn't be"
```
roman-khimov, line 866: should be?

pytest_tests/tests/acl/test_eacl_filters.py
```diff
@@ -627,3 +657,556 @@ def test_extended_acl_allow_filters_object(
...
```
roman-khimov, line 909: And here.

pytest_tests/tests/acl/test_eacl_filters.py
```diff
@@ -627,3 +657,556 @@ def test_extended_acl_allow_filters_object(
...
```
roman-khimov, line 814: should be

pytest_tests/tests/acl/test_eacl_filters.py
```diff
@@ -627,3 +657,556 @@ def test_extended_acl_allow_filters_object(
...
+        with allure.step("Add test object to container"):
+            file_path = generate_file(complex_object_size)
+
+            oid = put_object_to_random_node(
+                user_wallet.wallet_path,
+                file_path,
+                cid,
+                shell=self.shell,
+                neofs_env=self.neofs_env,
+                lifetime=2,
```
roman-khimov, line 993: I think you should use `expire-at`, lifetime is relative to the current epoch which technically can change during the test.

pytest_tests/tests/acl/test_eacl_filters.py
```diff
@@ -627,3 +657,556 @@ def test_extended_acl_allow_filters_object(
...
+                user_wallet.wallet_path,
+                cid,
+                shell=self.shell,
+                endpoint=self.neofs_env.sn_rpc,
+            )
+
+        small_file_path = generate_file(complex_object_size)
+
+        with allure.step(
+            f"PUT object with should not be allowed because size is LT {complex_object_size + 1}"
```
roman-khimov, line 1071: s/with/which/

pytest_tests/tests/acl/test_eacl_filters.py
evgeniiz321: fixed

pytest_tests/tests/acl/test_eacl_filters.py
evgeniiz321: fixed

pytest_tests/tests/acl/test_eacl_filters.py
evgeniiz321: fixed

pytest_tests/tests/acl/test_eacl_filters.py
evgeniiz321: fixed

pytest_tests/tests/acl/test_eacl_filters.py
evgeniiz321: fixed

roman-khimov (approved):

PR #766, opened 2024-03-29
tests: add tests for search api with numeric attrs

### LINKED ISSUE #739: Test numeric SEARCH queries
We can SEARCH using numeric comparisons since https://github.com/nspcc-dev/neofs-node/pull/2733 (see https://github.com/nspcc-dev/neofs-api/issues/265 also), this needs to be tested for small/big objects with positive/negative scenarios.

evgeniiz321:
closes #739

### DIFF
--- a/pytest_tests/lib/helpers/grpc_responses.py
+++ b/pytest_tests/lib/helpers/grpc_responses.py
@@ -40,6 +40,8 @@ EACL_PROHIBITED_TO_MODIFY_SYSTEM_ACCESS = (
     "table validation: it is prohibited to modify system access"
 )
 
+INVALID_SEARCH_QUERY = ".*invalid search query.*"
+
 
 def error_matches_status(error: Exception, status_pattern: str) -> bool:
     """
--- a/pytest_tests/tests/object/test_object_api.py
+++ b/pytest_tests/tests/object/test_object_api.py
@@ -18,6 +18,7 @@ from helpers.grpc_responses import (
     INVALID_OFFSET_SPECIFIER,
     INVALID_RANGE_OVERFLOW,
     INVALID_RANGE_ZERO_LENGTH,
+    INVALID_SEARCH_QUERY,
     OUT_OF_RANGE,
 )
 from helpers.neofs_verbs import (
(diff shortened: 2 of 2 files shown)


pytest_tests/tests/object/test_object_api.py
```diff
@@ -52,6 +53,16 @@
 RANGE_MAX_LEN = 500
 # Used for static ranges found with issues
 STATIC_RANGES = {}
+OBJECT_NUMERIC_VALUES = [-1, 0, 1, 10]
```
roman-khimov, line 56: We need some `xyz` value as well to check that this object is never returned for numeric searches (attributes can have any values).

Some large (>2^64) values would also be nice to check.

evgeniiz321: Do you mean this filter `attr GT xyz`? if we have letters instead of numbers we got `invalid search query` error, I have a test for that. 

For large value - added it.

roman-khimov: I mean searching for `attr GT 0`, but having an object with `attr=xyz`.

pytest_tests/lib/helpers/grpc_responses.py
```diff
@@ -40,6 +40,8 @@
     "table validation: it is prohibited to modify system access"
 )
 
+INVALID_SEARCH_QUERY = ".*invalid search query.*"
```
cthulhu-rider, line 43: i'd also add query(s) with floating point numbers which are also currently unsupported. `attr<1.2` is incorrect as well as `attr<text` but not so obvious

pytest_tests/tests/object/test_object_api.py
evgeniiz321: added both scenarios

pytest_tests/lib/helpers/grpc_responses.py
evgeniiz321: added this case as well

pytest_tests/tests/object/test_object_api.py
```diff
@@ -281,6 +292,133 @@ def _generate_filters_expressions(attrib_dict: dict[str, str]):
...
+    def test_object_search_with_numeric_queries(
+        self, default_wallet: NodeWallet, container: str, object_size: int, operator: str
+    ):
+        objects = []
+        for numeric_value in OBJECT_NUMERIC_VALUES:
+            file_path = generate_file(object_size)
+
+            objects.append(
+                {
+                    "numeric_value": numeric_value,
```
cthulhu-rider, line 310: `"numeric_value"` key string used multiple times, can we define a const for it?

cthulhu-rider (approved):

pytest_tests/tests/object/test_object_api.py
evgeniiz321: Thank you, there was already a const for it, i just didn't use it here. Fixed.

roman-khimov (approved):

PR #770, opened 2024-03-29
Drop HTTP gateway from tests, closes #757

### LINKED ISSUE #757: Drop HTTP gateway from tests
I'm always frustrated when we're testing deprecated software. HTTP gateway is gone, so we don't need to waste time testing it.

Ensure we have no HTTP-specific tests, drop it from tests that run both for HTTP and REST gateways (testing REST gateway only now).

evgeniiz321:


pytest_tests/pytest.ini
```diff
@@ -15,8 +15,7 @@ markers =
     grpc_api: standard gRPC API tests
     grpc_control: tests related to using neofs-cli control commands
     grpc_object_lock: gRPC lock tests
-    http_and_rest_gates: HTTP and REST gates contract
-    http_gate: HTTP gate contract
+    rest_gate: REST gates contract
```
roman-khimov, line 18: Contract?

evgeniiz321: Fixed

roman-khimov (approved):

PR #777, opened 2024-04-11
tests: fix rest gw tests

evgeniiz321:


pytest_tests/tests/services/rest_gate/test_rest_headers.py
```diff
@@ -217,13 +217,12 @@ def test_negative_put_and_get_object3(
         with allure.step(
             "[Negative] Try to download (wget) object via wget with attributes [peace=peace]"
         ):
-            request = f"/get/{storage_object_1.cid}/peace/peace"
```
roman-khimov, line 220: But why? I thought it should be a proper request to `/get_by_attribute/` that fails because of missing container.

pytest_tests/tests/services/rest_gate/test_rest_gate.py
```diff
@@ -335,11 +335,10 @@ def test_put_http_get_http(self, complex_object_size, simple_object_size, gw_par
     @pytest.mark.parametrize(
         "attributes",
         [
-            {"fileName": "simple_obj_filename"},
-            {"file-Name": "simple obj filename"},
-            {"cat%jpeg": "cat%jpeg"},
+            {"Filename": "simple_obj_filename"},
```
roman-khimov, line 338: FileName, it's special.

pytest_tests/tests/services/rest_gate/test_rest_gate.py
```diff
@@ -335,11 +335,10 @@ def test_put_http_get_http(self, complex_object_size, simple_object_size, gw_par
     @pytest.mark.parametrize(
         "attributes",
         [
-            {"fileName": "simple_obj_filename"},
-            {"file-Name": "simple obj filename"},
-            {"cat%jpeg": "cat%jpeg"},
```
roman-khimov, line 340: You're removing it without a notice, it should be a separate commit. https://github.com/nspcc-dev/neofs-rest-gw/issues/195 is the real reason (that we can't fix now).

roman-khimov (changes requested):

pytest_tests/tests/services/rest_gate/test_rest_headers.py
evgeniiz321: You are right, inside `try_to_get_object_via_passed_request_and_expect_error` there is a call to `/get_by_attribute/` and it works as expected, but if we pass the request string directly to the method it uses it, but there is no sense in it, since we pass attritbutes there. This test just has never worked properly.

pytest_tests/tests/services/rest_gate/test_rest_gate.py
evgeniiz321: I thought it was an officially  unsupported symbol in current rest gw, returned it back and skipped properly.

pytest_tests/tests/services/rest_gate/test_rest_gate.py
evgeniiz321: I added a separate case for FileName

roman-khimov (approved):

PR #790, opened 2024-04-26
Bunch of tiny fixes for macos

evgeniiz321:


pytest_tests/lib/helpers/http_gate.py
```diff
@@ -222,7 +222,7 @@ def upload_via_http_gate_curl(
     large_object = is_object_large(filepath)
     if large_object:
         # pre-clean
-        _cmd_run("rm pipe -f")
+        _cmd_run("rm pipe -f || true")
```
roman-khimov, line 225: dangerous

roman-khimov: Maybe we can make something more darwin-specific here?

evgeniiz321: why is it dangerous? we just ignore the RC here, if something is not ok, we will fail on the next command. I'm not sure it is worth it to adapt the code to macos here, since the problem is only in the `rm -f` returning non zero RC on macos if there is nothing to remove. But if we want, I can try to get rid of pipes here at all

roman-khimov: It can fail for some other reason and you won't notice. At least if the problem doesn't affect the next command. But you're right in that `mkfifo` is the main thing here, if it works then we don't care much about  `rm`. Does it all happen in tmp directory?

evgeniiz321: Nope, it happens in the current working directory.

roman-khimov: That's worse. Maybe we can feed `curl` in some other way then or use some pythonic library instead of curl. This named pipe trick came from the HTTP gateway documentation. There CLI tooling only was used, so there were not a lot of choices. But in a test it seems kludgy, this pipe can be left in FS if something goes wrong.

PR #793, opened 2024-05-02
Fix/object parts changes

carpawell:


roman-khimov (commented):
Do we have any test for allowed deletions (incomplete objects)?

carpawell:
> Do we have any test for allowed deletions (incomplete objects)?

No. I am not sure if it can be done without additional support in our utils. For now we can only put big object and stop after N seconds?

roman-khimov:
Let's create an issue and think about it separately then.

carpawell:
@roman-khimov, test repo or cli? I agree such a test will be helpful, but do not see any easy solution without some part creation util.

roman-khimov:
Test. We'll untangle from it.

roman-khimov (approved):

evgeniiz321 (approved):

PR #798, opened 2024-05-07
Bunch of fixes for macos

evgeniiz321:


pytest_tests/lib/helpers/http_gate.py
```diff
@@ -212,24 +221,9 @@ def upload_via_http_gate_curl(
...
-        cmd = f"mkfifo pipe;cat {filepath} > pipe & curl --silent --no-buffer -F '{files}' {attributes}{cookies_attr} {request}"
-        output = _cmd_run(cmd, LONG_TIMEOUT)
-        # clean up pipe
-        _cmd_run("rm pipe")
-    else:
-        files = f"file=@{filepath};filename={os.path.basename(filepath)}"
-        cmd = f"curl --silent -F '{files}' {attributes}{cookies_attr} {request}"
-        output = _cmd_run(cmd)
+    files = f"file=@{filepath};filename={os.path.basename(filepath)}"
+    cmd = f"curl --silent -F '{files}' {attributes} {request}"
```
roman-khimov, line 225: I'm not sure it's equivalent. The pipe was used to test streaming upload with unknown file size. IIRC in this case `curl` will add size to the request immediately (because it's known, unlike when you're reading data from a pipe).

evgeniiz321: This is not equivalent, curl is not used to upload files anymore. Basically, curl is only needed for just one test - to verify rest gw behavior when duplicated headers exist in the request. Surprisingly python libraries can't do that easy way - it is easier to use curl in this case.

roman-khimov (approved):

pytest_tests/lib/helpers/http_gate.py
roman-khimov: #799

PR #808, opened 2024-06-07
tests: save temp files in case of errors

evgeniiz321:


evgeniiz321:
@roman-khimov in case of failures, we attach two archives with env files ~250MB and with temp files ~600MB. Is it ok? If it is, then we can merge, if there are no other comments.

roman-khimov:
We push them into NeoFS, it's OK.

roman-khimov (approved):

PR #815, opened 2024-06-16
tests: enable metadata failover tests

evgeniiz321:


evgeniiz321:
will be done in https://github.com/nspcc-dev/neofs-testcases/pull/814

PR #826, opened 2024-06-28
tests: add checks for object parts removal

### LINKED ISSUE #809: Add check that object parts are removed from storages on expiration
If a big object is removed, all its parts should be also dropped. There was https://github.com/nspcc-dev/neofs-node/issues/2858 issue and https://github.com/nspcc-dev/neofs-node/pull/2863 solution for it. It seemed that it was never tested and garbage is always near big objects with expiration.

1. Create a big object
2. Get (remember) its parts
3. Expire/Delete it (both are good to check)
4. Try to get its parts with public API (it would also be good to check with `neofs-lens` that storage it cleared physically, not only marked as unavailable)

evgeniiz321:
closes #809


evgeniiz321:
@carpawell can you please take a look. And also can you please send an example command from neofs-lens to inspect the cleared storage after objects deletion?

roman-khimov (approved):

PR #829, opened 2024-07-24
tests: use binaries from local dir

### LINKED ISSUE #398: NeoFS binaries should be installed in venv
Testcases require some NeoFS and Neo binaries available. It would be nice to install them inside virtualenv instead of using /usr/local/bin, as stated in README.

Maybe we could also pin a particular versions and automate the environment construction as it's done in neofs-dev-env, for example.

evgeniiz321:
closes #398


roman-khimov (commented):
Looks legit

roman-khimov (approved):

PR #831, opened 2024-08-03
env: add version dependent config for s3 gw

evgeniiz321:


.github/workflows/run-tests.yml
```diff
@@ -237,7 +237,7 @@ jobs:
         env:
           ALLURE_RESULTS_DIR: ${{ env.ALLURE_RESULTS_DIR }}
         run: |
-          source venv.pytest/bin/activate && pytest --alluredir=${GITHUB_WORKSPACE}/allure-results pytest_tests/tests
+          source venv.pytest/bin/activate && pytest --log-cli-level error --alluredir=${GITHUB_WORKSPACE}/allure-results pytest_tests/tests
```
roman-khimov, line 240: Why?

evgeniiz321: We don't need all of the logs when running in CI. They make it harder to see actual errors. Besides, we regularly have allure, that makes these logs even more useless in CI. When running locally - it is a different story and the log level is INFO there.

roman-khimov: Why is it not explained in the commit message?

evgeniiz321: Updated the commit message.

roman-khimov (approved):

PR #836, opened 2024-08-13
tests: add storage nodes restart to payment tests

evgeniiz321:


pytest_tests/lib/helpers/node_management.py
```diff
@@ -68,14 +68,19 @@ def stop_storage_nodes(nodes: list[StorageNode]) -> None:
...
+    The function restarts all storage nodes of the specified neofs env.
     Args:
-       nodes: the list of nodes to restart
+       neofs_env: NeoFSEnv instance
     """
-    stop_storage_nodes(nodes)
-    start_storage_nodes(nodes)
+    for node in neofs_env.storage_nodes:
+        node.stop()
+    time.sleep(5)
```
roman-khimov, line 79: Doesn't seem to be reliable.

pytest_tests/lib/helpers/node_management.py
```diff
@@ -68,14 +68,19 @@ def stop_storage_nodes(nodes: list[StorageNode]) -> None:
...
     """
-    The function restarts specified storage nodes.
+    The function restarts all storage nodes of the specified neofs env.
     Args:
-       nodes: the list of nodes to restart
+       neofs_env: NeoFSEnv instance
     """
-    stop_storage_nodes(nodes)
-    start_storage_nodes(nodes)
+    for node in neofs_env.storage_nodes:
```
carpawell, line 77: can you please add some context to the commit? it is not clear to me why we need it and, moreover, why we cannot use `stop_storage_nodes(neofs_env.storage_nodes)`

pytest_tests/lib/helpers/node_management.py
evgeniiz321: replaced with an explicit wait for the process termination

pytest_tests/lib/helpers/node_management.py
evgeniiz321: Fixed. The only needed change here was to update `start_storage_nodes` function.

roman-khimov: It's not obvious from the commit message _why_ we're changing this. And it's just a workaround, not something critical to the test logic, so this better be explained for people reading the code in a year.

roman-khimov (approved):

PR #844, opened 2024-08-23
tests: update existing eacl tests to add usage of an address as a target

### LINKED ISSUE #838: Cover account-based EACL targets
I'm always frustrated when we don't have any tests for cool new features we deliver. https://github.com/nspcc-dev/neofs-api/issues/278 is implemented in the node and it must be tested.

Basically, anywhere a key can be used in EACLs an address can be used now as well. So we need to extend our test suite to check for that.

evgeniiz321:
closes #838

### DIFF
--- a/pytest_tests/lib/helpers/acl.py
+++ b/pytest_tests/lib/helpers/acl.py
@@ -15,6 +15,7 @@ from helpers.data_formatters import get_wallet_public_key
 from helpers.grpc_responses import EACL_NOT_FOUND, EACL_TABLE_IS_NOT_SET
 from neofs_testlib.cli import NeofsCli
 from neofs_testlib.shell import Shell
+from neofs_testlib.utils.wallet import get_last_address_from_wallet
 
 logger = logging.getLogger("NeoLogger")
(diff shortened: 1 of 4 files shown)


pytest_tests/lib/helpers/acl.py
```diff
@@ -42,6 +43,26 @@ class EACLRole(Enum):
     SYSTEM = "system"
 
 
+class EACLRoleExtendedType(Enum):
+    PUBKEY = "pubkey"
+    ADDRESS = "address"
+
+    def get_value(self, wallet_path: str, wallet_password: str) -> str:
```
roman-khimov, line 50: Why `wallet_password` here? Technically you don't need it, public key is _public_, address is available as well.

evgeniiz321: It goes down to the APIs of the neo3 python package which requires a wallet password to be specified to even parse a wallet json file. I can implement my own methods to work with it, but not sure if it is worth it.

roman-khimov (approved):

PR #849, opened 2024-09-10
Enable pprof and prometheus for all services

### LINKED ISSUE #847: Enable pprof, metrics everywhere
Tests without availability to inspect them?..

## Current Behavior
No pprof, metrics enabled.

## Expected Behavior
Pprof, metrics enabled.

## Possible Solution
Chain: https://github.com/nspcc-dev/neofs-testcases/blob/a83d29566191799b3cbc8346c161e8516a81c353/neofs-testlib/neofs_testlib/env/templates/main_chain.yaml#L36-L39
FSChain: https://github.com/nspcc-dev/neofs-node/blob/d214b0c16e1a2d7ba692646e78ea5b1d23c6a973/config/example/ir.yaml#L194-L202
SN: https://github.com/nspcc-dev/neofs-testcases/blob/a83d29566191799b3cbc8346c161e8516a81c353/neofs-testlib/neofs_testlib/env/templates/sn.yaml#L8
S3: https://github.com/nspcc-dev/neofs-s3-gw/blob/55f9e7a15269f6b263aa4b5d0cefff0030921310/config/config.yaml#L56-L59
(however it is turned on for REST)

## Context
Never thought they are disabled. Understand that this information cannot be seen in regular CI but consider it as a disadvantage. In every failed run report may have a profile/metrics dump.
I have just run tests with `--persist-env` for ~30 mins and understood that I cannot inspect them cause profilers are turned off, need another 30 mins.

evgeniiz321:
closes #847

### DIFF
--- a/neofs-testlib/neofs_testlib/env/env.py
+++ b/neofs-testlib/neofs_testlib/env/env.py
@@ -542,6 +542,8 @@ class MainChain:
         self.main_chain_boltdb = self.neofs_env._generate_temp_file(extension="db", prefix="main_chain_bolt_db")
         self.rpc_address = f"{self.neofs_env.domain}:{NeoFSEnv.get_available_port()}"
         self.p2p_address = f"{self.neofs_env.domain}:{NeoFSEnv.get_available_port()}"
+        self.pprof_address = f"{self.neofs_env.domain}:{NeoFSEnv.get_available_port()}"
+        self.prometheus_address = f"{self.neofs_env.domain}:{NeoFSEnv.get_available_port()}"
         self.stdout = "Not initialized"
         self.stderr = "Not initialized"
         self.process = None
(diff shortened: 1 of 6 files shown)


carpawell (commented):
@evgeniiz321, is it possible to add them to the `env_details` file too? I use it a lot. Also since we are talking about it, it would be nice to make it more standard: for main services it has "RPC", "Endpoint" or "Address" names for the same things in fact. A control port is named "GRPC address" or "Control gRPC endpoint" depending on the service. Can create another issue if necessary.

evgeniiz321:
@carpawell updated env_details, addresses will be displayed there as well. As for standardization - pls open another issue.

roman-khimov (approved):

carpawell (approved):

PR #853, opened 2024-09-17
Simplify venv installation, #closes 676

evgeniiz321:


Makefile
```diff
@@ -1,28 +1,30 @@
 #!/usr/bin/make -f
 
-.DEFAULT_GOAL := help
+.DEFAULT_GOAL := venv.pytest
```
roman-khimov, line 3: You don't need it then.

.gitignore
```diff
@@ -19,7 +19,6 @@ xunit_results.xml
 
 # ignore work directories and setup files
 .setup
-.env
```
roman-khimov, line 22: Don't we use this file for local overrides?

Makefile
evgeniiz321: removed

.gitignore
evgeniiz321: In this repo - no, this file is not used and there is no code that handles it. Probably it was used in the past, but not now.

roman-khimov (approved):

cthulhu-rider (commented):
#closes XXX -> closes #XXX

PR #855, opened 2024-09-19
tests: check eACL can't be set with the final bit

### LINKED ISSUE #234: Add test to check if eACL cannot be set without resp. bit in container policy
Add a test to check that eACL cannot be set without the corresponding bit in the container's policy:

- create a container with public-read-write and separately arbitrary with Final-bit = 1
- try to set eACL - operation should not be performed
- check via neofs-cli container get-eacl that eACL is not present on the container.

evgeniiz321:
closes #234

### DIFF
--- a/pytest_tests/lib/helpers/grpc_responses.py
+++ b/pytest_tests/lib/helpers/grpc_responses.py
@@ -39,6 +39,8 @@ EACL_TIMED_OUT = "eACL setting: await timeout expired"
 EACL_TABLE_IS_NOT_SET = "extended ACL table is not set for this container"
 EACL_NOT_FOUND = "code = 3073.*message = eACL not found"
 EACL_PROHIBITED_TO_MODIFY_SYSTEM_ACCESS = "table validation: it is prohibited to modify system access"
+EACL_CHANGE_PROHIBITED = ".*container ACL is immutable.*"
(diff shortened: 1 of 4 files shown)


pytest_tests/lib/helpers/grpc_responses.py
```diff
@@ -39,6 +39,7 @@
 EACL_TABLE_IS_NOT_SET = "extended ACL table is not set for this container"
 EACL_NOT_FOUND = "code = 3073.*message = eACL not found"
 EACL_PROHIBITED_TO_MODIFY_SYSTEM_ACCESS = "table validation: it is prohibited to modify system access"
+EACL_CHANGE_PROHIBITED = ".*container ACL is immutable.*"
```
cthulhu-rider, line 42: this is thrown by CLI, so i suggest to also test behavior with `-f` flag disabling CLI-side pre-checks. Currently, i suppose exception wont be thrown in this case, and IR will deny such request async. Thus, timeout is expected

evgeniiz321: added a check with `-f` flag. Timeout is returned as you expected.

roman-khimov (approved):

cthulhu-rider (approved):

PR #861, opened 2024-09-25
tests: decrease timeout for a contract cache

### LINKED ISSUE #859: EACL tests take too much time
## Current Behavior
test_extended_acl_numeric_values --- each takes 3m30s to run.
test_extended_acl_numeric_values_attr_str_filter_numeric --- 30s
test_extended_acl_numeric_values_payload_attr --- 30s

https://rest.fs.neo.org/HXSaMJXk2g8C14ht8HSi7BBaiYZ1HeWh2xnWPGQCg4H6/544-1727128149/index.html#suites/7d0d2bff4573ced7a233aef01d2141a2/b20b4234dc3e1081/

## Expected Behavior
EACL is managed via blockchain transactions, but the chain we have in test environment has 1s blocks, so it does not take 15-30s to update anything there. I'd bet there is some excessive waiting in the test.

## Possible Solution
Try to optimize it.

evgeniiz321:
closes #859

### DIFF
--- a/pytest_tests/lib/helpers/acl.py
+++ b/pytest_tests/lib/helpers/acl.py
@@ -19,7 +19,7 @@ from neofs_testlib.utils.wallet import get_last_address_from_wallet
 
 logger = logging.getLogger("NeoLogger")
 EACL_LIFETIME = 100500
-NEOFS_CONTRACT_CACHE_TIMEOUT = 30
+NEOFS_CONTRACT_CACHE_TIMEOUT = 2
 
 
 class EACLOperation(Enum):


evgeniiz321:
@roman-khimov do you know what is this NEOFS_CONTRACT_CACHE_TIMEOUT and why are we waiting 30s for it after we set eacl? I decreased it to 2, all tests still pass, not sure if we need it at all.

roman-khimov:
Seems like it was based on ACL cache lifetime. But I can't find it now in the node. It also looks a lot like two 15 blocks, maybe related, maybe not. @cthulhu-rider?

cthulhu-rider:
cache timeout can be [configured](https://github.com/nspcc-dev/neofs-node/blob/1a9784c11ecc9f41193a4b99c5171ad470afb016/config/example/node.yaml#L75-L77). With [this](https://github.com/nspcc-dev/neofs-node/commit/ca8dc872b28e90d2cc66f9a28cf1c601da77ed96), defaullt expiration became 1 block. [Initially](https://github.com/nspcc-dev/neofs-node/commit/f9342aa562efc4bdef235e38306f117dd9625672), it was `30s`. I bet previous value was set back then

2s is pretty doubtful except for quick blocks <=1s. I see https://github.com/nspcc-dev/neofs-testcases/blob/f120678761c0957c8acf43144c0e03e70b89b201/pytest_tests/lib/helpers/acl.py#L176. If eACLs are updated with `--await` flag, having cache timeout is questionable. Only if the setting and reading are done from different endpoints

roman-khimov:
Test env uses 1s blocks, with `--await` it should OK to wait for cluster sync in 1-2 blocks, so 2s seems to be safe.

roman-khimov (approved):

PR #865, opened 2024-09-26
Bump python requirements to latest versions

evgeniiz321:


neofs-testlib/neofs_testlib/env/env.py
```diff
@@ -1152,7 +1152,7 @@ def prepare(
         size=1024,
         containers=1,
         out="grpc.json",
-        preload_obj=250,
+        preload_obj=100,
```
cthulhu-rider, line 1155: can u explain meaning of this value? If as u said `creating a high number of objects doesn't make sense`, it can be `10` then?

pytest_tests/tests/services/rest_gate/test_rest_system_header.py
```diff
@@ -9,16 +9,16 @@
...
     attr_into_str_header,
     try_to_get_object_and_expect_error,
     upload_via_rest_gate,
 )
-from helpers.neofs_verbs import get_netmap_netinfo, get_object_from_random_node, head_object
 from helpers.wellknown_acl import PUBLIC_ACL
-from rest_gw.rest_utils import get_object_and_verify_hashes
 from neofs_env.neofs_env_test_base import NeofsEnvTestBase
 from pytest_lazy_fixtures import lf
+from rest_gw.rest_utils import get_object_and_verify_hashes
```
roman-khimov, line 21: More like formatting rather than optimization (wrt commit message).

neofs-testlib/neofs_testlib/env/env.py
roman-khimov: What's the difference between 250 and 100 in practice?

evgeniiz321: 1m30s to create 250 objects and 30s to create 100 objects. 1m30s can cause some timeouts (which I encountered in this PR) inside tests. So I decreased it a little bit. 10 objects is too small as for me for this scenario, I feel like we can miss something with 10, but this is just feelings. I can decrease up to 2 objects actually.

pytest_tests/tests/services/rest_gate/test_rest_system_header.py
evgeniiz321: updated commit message

neofs-testlib/neofs_testlib/env/env.py
roman-khimov: OK for me, but you need to explain this in the commit message.

evgeniiz321: updated the commit message

roman-khimov (approved):

PR #867, opened 2024-09-30
tests: tick epoch after its duration change

evgeniiz321:


pytest_tests/tests/payment/test_container_payments.py
```diff
@@ -165,8 +167,9 @@ def test_container_payments(
                 ):
                     storage_nodes_info[sn]["objects_count"] += 1
 
-        with allure.step("Wait until a new epoch arrives"):
+        with allure.step("Wait for a couple of epochs to arrive"):
             new_epoch = neofs_epoch.wait_until_new_epoch(neofs_env, neofs_epoch.get_epoch(neofs_env))
+            new_epoch = neofs_epoch.wait_until_new_epoch(neofs_env, new_epoch)
```
carpawell, line 172: it does not work without it?

carpawell (approved):

pytest_tests/tests/payment/test_container_payments.py
roman-khimov: I think this gets us a complete clean epoch with objects present in the container. We upload in the previous one, so it can be racy wrt estimations.

roman-khimov (approved):

cthulhu-rider (approved):
commit text looks a bit strange

pytest_tests/tests/payment/test_container_payments.py
evgeniiz321: As @roman-khimov said, without it, we sometimes _can_ get not accurate values that we can't compare later and it causes test to fail.

PR #881, opened 2024-10-22
.github: use common system tests template

evgeniiz321:


roman-khimov:
Looks like this never ends:

>  run_system_tests Expected — Waiting for status to be reported

evgeniiz321:
@roman-khimov i assume this is because github is waiting for a single job here, and i added a matrix with macos. Need to update the repo settings

roman-khimov:
It allows to add any of subordinate jobs, but also works with top-level.

roman-khimov:
Let's check what's gonna happen after the merge.

PR #890, opened 2024-10-31
.github: add timeout for individual tests + bump dependencies

evgeniiz321:


evgeniiz321:
@roman-khimov can you please check repo settings once again? Maybe we need to remove and add once again the rule for run_system_tests job.

roman-khimov:
Deleted/added back. Let's see how it goes.

evgeniiz321:
@roman-khimov I tried setting rules on my demo repo and I can select individual matrix jobs there:

image

can you do the same here?

roman-khimov:
Of course I can. But it allows to select parent entity as well which doesn't work. Weird.

.github/workflows/system-tests.yml
```diff
@@ -112,20 +112,20 @@ jobs:
         timeout-minutes: 500
         steps:
           - name: Set up Go
-            uses: actions/setup-go@v5
+            uses: actions/setup-go@v5.1.0
```
roman-khimov, line 115: This shouldn't be needed, v5 tracks the latest v5.X.Y.

.github/workflows/system-tests.yml
```diff
@@ -180,7 +180,7 @@ jobs:
 
           - name: Download neofs-cli
             if: ${{ inputs.neofs_node_tag != 'from_tests' }}
-            uses: dsaltares/fetch-gh-release-asset@1.1.1
+            uses: dsaltares/fetch-gh-release-asset@1.1.2
```
roman-khimov, line 183: Does it have a `v1` of some kind?

.github/workflows/system-tests.yml
```diff
@@ -372,7 +372,7 @@ jobs:
             if: always() && steps.put_report.outcome == 'success'
             env:
               REPORT_NEOFS_URL: ${{ steps.put_report.outputs.REPORT_NEOFS_URL }}index.html
-            uses: Sibz/github-status-action@v1
+            uses: guibranco/github-status-action-v2@v1.1.13
```
roman-khimov, line 375: That's different. Are we sure it's OK to use?

.github/workflows/system-tests.yml
roman-khimov: Same applies to other @vX

.github/workflows/system-tests.yml
evgeniiz321: the old one is deprecated

.github/workflows/system-tests.yml
evgeniiz321: Nope

.github/workflows/system-tests.yml
roman-khimov: Looks like not a lot of choices at the moment.

roman-khimov (approved):

PR #892, opened 2024-11-04
tests: verify SN can be stopped using SIGTERM

evgeniiz321:


pytest_tests/tests/failovers/test_failover_storage.py
```diff
@@ -56,7 +56,7 @@ def test_storage_node_failover(
             if hard_restart:
                 node_to_stop.kill()
             else:
-                node_to_stop.stop()
+                node_to_stop.stop(force=False)
```
roman-khimov, line 59: But if it fails, we're stuck here, right? Maybe it's better to pass/check some return code when node is killed.

evgeniiz321: no, if it fails - we will get a TimeoutError after 60 seconds

evgeniiz321: flag is about what to do when we get this TimeoutError - with force - we use kill, without force - we raise an error

evgeniiz321: @roman-khimov i adjusted this a little, but the main idea is the same - if a node doesn't stop gracefully - we kill it, and fail the test but no hangs/stuck behavior anywhere.

roman-khimov (approved):

PR #894, opened 2024-11-08
tests: adapt s3 acl tests to new default behavior

### LINKED ISSUE #885: S3 ACL default behavior test
I'm always frustrated when we don't have any tests for https://github.com/nspcc-dev/neofs-s3-gw/issues/1016.

Add 0.32.1+ test that will:
 * check the default ACL
 * check PUTs without ACL and with owner-control ACL (many of these, trying to reproduce EACL bloat which can be tested with S3 0.32.0, it's degrading quickly)
 * check PUTs with other ACLs that MUST fail unless we enable ACLs explicitly

This intersects with S3 testcases and s3-tests we already have.

evgeniiz321:
closes #885

### DIFF
--- a/pytest_tests/lib/helpers/aws_cli_client.py
+++ b/pytest_tests/lib/helpers/aws_cli_client.py
@@ -534,6 +534,15 @@ class AwsCliClient:
         output = _cmd_run(cmd)
         return self._to_json(output)
 
+    def put_bucket_ownership_controls(self, Bucket, OwnershipControls):
+        cmd = (
+            f"aws {self.common_flags} s3api put-bucket-ownership-controls --bucket {Bucket} "
+            f"--ownership-controls '{{\"Rules\": [{{\"ObjectOwnership\": \"{OwnershipControls['Rules'][0]['ObjectOwnership']}\"}}]}}' "
+            f"--endpoint-url {self.s3gate_endpoint}"
+        )
+        output = _cmd_run(cmd)
+        return self._to_json(output)
+
     @staticmethod
     def _to_json(output: str) -> dict:
         json_output = {}
--- a/pytest_tests/lib/s3/s3_bucket.py
+++ b/pytest_tests/lib/s3/s3_bucket.py
@@ -23,6 +23,12 @@ class VersioningStatus(Enum):
     SUSPENDED = "Suspended"
 
 
+class ObjectOwnership(Enum):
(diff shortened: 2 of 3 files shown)


evgeniiz321:
@roman-khimov shouldn't it be 2 different operations to enable ACLs for objects and buckets? Right now this command:
```
s3_bucket.put_bucket_policy(
                    self.s3_client,
                    bucket,
                    {
                        "Statement": [
                            {
                                "Sid": "BucketEnableACL",
                                "Effect": "Allow",
                                "Action": ["s3:PutObject"],
                                "Resource": ["*"],
                            }
                        ],
                    },
                )
  ```
  Enables ACLs both for objects and ACLs, another thing - only this Action is accepted - `"Action": ["s3:PutObject"],`, all other causes 500 from s3 gw.

roman-khimov:
I think it's better to be verified against Amazon. But I've also found https://github.com/nspcc-dev/neofs-s3-gw/issues/1023 and https://github.com/nspcc-dev/neofs-s3-gw/issues/1022.

evgeniiz321:
@smallhive why are we requiring this: `header x-amz-acl:bucket-owner-full-control must be set` ? neither boto3 client nor aws s3 cli set this header. Amazon also doesn't require it, see - https://github.com/nspcc-dev/neofs-s3-gw/issues/1023#issuecomment-2469375817

smallhive:
> header x-amz-acl:bucket-owner-full-control must be set

It should be valid, in the `preferred` mode only. In `object writer`, it should work fine with any ACLs.
I did this according to [docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/ensure-object-ownership.html#ensure-object-ownership-bucket-policy), but it works without the bucket policy (this policy requires bucket-owner-full-control ACL), described in the docs

evgeniiz321:
@smallhive @roman-khimov this is different from the amazon behavior. With BucketOwnerPreferred, any object ACLs can be set, if there are no other policies. _But_ if a bucket owner wants, there is a possibility to add a policy to make object writers to forcefully specify bucket-owner-full-control. This is what [docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/ensure-object-ownership.html#ensure-object-ownership-bucket-policy) is about. This behavior shouldn't be enforced by default, but should be enabled via policies. If we want to behave the same as the amazon s3 implementation.

roman-khimov (approved):

PR #897, opened 2024-11-20
tests: container creation with --global-name

### LINKED ISSUE #138: Containers with nice names
Test containers with `__NEOFS_NAME` and `__NEOFS__ZONE` attributes. Check availability of container name in NNS contract (TXT record should return string with base58 encoded container ID).

Testcases:
1. Container created with `__NEOFS_NAME=foo` attribute available in NNS with `foo.container` domain.
2. Container with the same name can't be created twice.
3. Container can be created with name and `__NEOFS__ZONE=container`. 
4. Container can't be created with any other zone.
5. Container can' be created with zone but without name.

evgeniiz321:
closes #138

### DIFF
--- a/pytest_tests/tests/container/test_container.py
+++ b/pytest_tests/tests/container/test_container.py
@@ -211,3 +211,38 @@ class TestContainer(NeofsEnvTestBase):
 
         with allure.step("Previously stopped node should return GC MARKED for the created object"):
             object_should_be_gc_marked(self.neofs_env, node_to_stop, cid, oid)
+
+    def test_container_global_name(self, default_wallet, simple_object_size):
+        if self.neofs_env.storage_nodes[0]._get_version() <= "0.43.0":
+            pytest.skip("This test runs only on post 0.43.0 node version")
+
+        with allure.step("Create container"):
+            wallet = default_wallet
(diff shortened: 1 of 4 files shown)


evgeniiz321:
@roman-khimov in the original ticket https://github.com/nspcc-dev/neofs-testcases/issues/138, this attribute - `__NEOFS__ZONE` is also mentioned, do we want to have this attribute in the container create cli command?

roman-khimov:
I'd not bother, many things would break if we're to touch it.

roman-khimov (approved):

pytest_tests/tests/container/test_container.py
```diff
@@ -211,3 +211,35 @@ def test_container_deletion_while_sn_down(
 
         with allure.step("Previously stopped node should return GC MARKED for the created object"):
             object_should_be_gc_marked(self.neofs_env, node_to_stop, cid, oid)
+
+    def test_container_global_name(self, default_wallet, simple_object_size):
```
roman-khimov, line 215: This test should be version-dependent, it fails with 0.43.0.

evgeniiz321: fixed

PR #898, opened 2024-11-22
tests: remove locked object from non owner node

evgeniiz321:
@carpawell this is a scenario from https://github.com/nspcc-dev/neofs-testcases/issues/895#issuecomment-2478988417, but it doesn't reproduce the original problem on 0.43.0. Nevertheless, the scenario is worth to add in any case. But for the https://github.com/nspcc-dev/neofs-testcases/issues/895 I need some help.

### DIFF
--- a/neofs-testlib/neofs_testlib/cli/neofs_cli/container.py
+++ b/neofs-testlib/neofs_testlib/cli/neofs_cli/container.py
@@ -264,3 +264,25 @@ class NeofsCliContainer(CliCommand):
             "container set-eacl",
             **{param: value for param, value in locals().items() if param not in ["self"]},
         )
+
+    def nodes(
+        self,
+        rpc_endpoint: str,
+        wallet: str,
+        cid: str,
+    ) -> CommandResult:
+        """
+        Show nodes taking part in a container at the current epoch.
+
+        Args:
+            rpc_endpoint: Remote node address (as 'multiaddr' or '<host>:<port>').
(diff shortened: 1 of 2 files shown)


carpawell:
OK, i understood the reason: relay works and worked only for signed objects: https://github.com/nspcc-dev/neofs-node/blob/edc95f6a729d29d9fdaf6a54f4dfa5bc4d3b9c3c/pkg/services/object/put/streamer.go#L72-L73. Tombstone and any other non-regular object are ready and signed but, of course, they do not go through the relay process as it is turned off. So basically we are stuck here cause no easy public functionality for pushing already signed objects (too often we are thinking of https://github.com/nspcc-dev/neofs-node/issues/952). Why it happened in our test load runs: https://github.com/nspcc-dev/neo-go/blob/e89f9fe2a417c917d40d933c1b6a5f92804d6e2d/cli/util/upload_bin.go#L597-L600, still surprised why you went this way but that is how uploader sends prepared and signed objects. It is really not a critical case as I do not know who else is going to upload objects this way (and also we have fixed it just no auto test for now).
cc @roman-khimov

evgeniiz321:
@roman-khimov i would merge this PR, the test is valid and checks some interesting scenario. It doesn't validate https://github.com/nspcc-dev/neofs-testcases/issues/895, but still a good test case to have. For https://github.com/nspcc-dev/neofs-testcases/issues/895 we will create a new test.

carpawell (approved):
Agree as soon as it does not close the issue.

PR #903, opened 2024-12-03
env: use the latest and greatest NeoGo

roman-khimov:
0.107.0 defines Echidna, so it's important to have it.


evgeniiz321:
@roman-khimov IR with main chain fail with:
```
warn	client/multi.go:43	could not establish connection to RPC node	{"endpoint": "ws://localhost:52519/ws", "error": "WS client initialization: failed to get network magic: unexpected hardfork: Echidna"}
```
Does it require some config values to be updated?

roman-khimov:
Can be done with a config, can be done with an update to NeoFS. Looks like we can not do this upgrade cleanly, but step by step it'll be OK.

roman-khimov:
We need to move on with it. Some breakage is unavoidable, unfortunately.

PR #929, opened 2025-01-17
tests: add tests for IR controls for setting management

### LINKED ISSUE #924: Test IR controls for setting management
I'm always frustrated when we don't have tests for new functionality.

We've added new commands in https://github.com/nspcc-dev/neofs-node/pull/3059 and we need some test for them. Caveat: we need at least 4 IR nodes. Sending/listing/confirming and waiting for the outcome are all included.

evgeniiz321:
closes #924

### DIFF
--- a/pytest_tests/tests/network/test_ir_controls.py
+++ b/pytest_tests/tests/network/test_ir_controls.py
@@ -1,0 +1,226 @@
+import time
+from typing import Union
+
+import allure
+import neofs_env.neofs_epoch as neofs_epoch
+import pytest
+from helpers.neofs_verbs import CONFIG_KEYS_MAPPING, get_netmap_netinfo
+from helpers.test_control import wait_for_success
+from neofs_testlib.env.env import NeoFSEnv, StorageNode
+from neofs_testlib.utils import wallet as wallet_utils
+
+
+@allure.step("Wait until new config value arrives")
+@wait_for_success(60, 5)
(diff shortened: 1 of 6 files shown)


pytest_tests/tests/network/test_ir_controls.py
```diff
@@ -0,0 +1,226 @@
...
+        for ir_node in neofs_env.inner_ring_nodes[1:-1]:
+            neofs_env.neofs_cli(ir_node.cli_config).control.notary_sign(
+                address=ir_node.alphabet_wallet.address,
+                endpoint=ir_node.grpc_address,
+                wallet=ir_node.alphabet_wallet.path,
+                hash=tx_hash,
+            )
+
+    with allure.step("Wait until new epoch arrives"):
+        neofs_epoch.wait_until_new_epoch(neofs_env, current_epoch)
```
roman-khimov, line 114: This should limited in time to some low number of blocks, otherwise it can tick on its own.

evgeniiz321: `neofs_epoch.wait_until_new_epoch` waits for 60s before failing. There is a call `neofs_epoch.ensure_fresh_epoch(neofs_env)` above, and default epoch length == 240 is used. This should be enough to ensure it doesn't tick on its own

roman-khimov (approved):

PR #948, opened 2025-02-09
tests: add write cache test

evgeniiz321:


evgeniiz321:
https://github.com/nspcc-dev/neofs-node/issues/3114

pytest_tests/tests/network/test_deployment.py
```diff
@@ -92,3 +117,58 @@ def test_multiple_ir_node_deployment(ir_nodes_count: int, clear_neofs_env: NeoFS
...
+            },
+            writecache=True,
+        )
+
+        neofs_env.neofs_adm().fschain.set_config(
+            rpc_endpoint=f"http://{neofs_env.fschain_rpc}",
+            alphabet_wallets=neofs_env.alphabet_wallets_dir,
+            post_data="ContainerFee=0 ContainerAliasFee=0 MaxObjectSize=524288",
+        )
+        time.sleep(30)
```
roman-khimov, line 143: Why?

pytest_tests/tests/network/test_deployment.py
```diff
@@ -92,3 +117,58 @@ def test_multiple_ir_node_deployment(ir_nodes_count: int, clear_neofs_env: NeoFS
...
+                object_address = neofs_lens.write_cache.list(shard_wc_path).stdout.strip()
+                if object_address:
+                    object_info = parse_write_cache_get_data(
+                        neofs_lens.write_cache.get(object_address, shard_wc_path).stdout
+                    )
+                    assert object_info["CID"] == cid, "Invalid value in the write cache"
+                    assert object_info["ID"] == oid, "Invalid value in the write cache"
+                    assert int(object_info["PayloadSize"]) == size, "Invalid value in the write cache"
+                    shards_write_caches_with_object += 1
+        assert shards_write_caches_with_object == 3, (
```
roman-khimov, line 166: It should be in one shard and in fact it's racy, write cache is write cache, it will be flushed asap.

pytest_tests/tests/network/test_deployment.py
evgeniiz321: It was some rudimentary wait for applying network settings. Removed it. Works without it as well.

pytest_tests/tests/network/test_deployment.py
evgeniiz321: Updated naming to show these are shards from different nodes. And let's leave the racy part as is right now. When the behavior will be enhanced - will see a failure and come and fix it the right way.

pytest_tests/tests/network/test_deployment.py
roman-khimov: Still here.

evgeniiz321: Removed. + Added a part with neofs-lens storage commands on stopped nodes.

roman-khimov (approved):

PR #958, opened 2025-02-25
refactor env usage and fix known issues

evgeniiz321:
Move the module scoped vars to function scoped vars, since module scoped vars do not work correctly when running in multithreaded/multiprocessed environments. It caused other tests to fail in case of errors inside test_container_deletion_while_sn_down.


roman-khimov:
A number of tests fail here.

evgeniiz321:
latest failures are not related to the changes in this PR

roman-khimov (approved):

PR #964, opened 2025-02-27
Add asserts for zero ranges

evgeniiz321:


pytest_tests/tests/object/test_object_api.py
```diff
@@ -562,6 +562,28 @@ def test_object_get_range_hash(
                         f"Expected range hash to match {range_cut} slice of file payload"
                     )
 
+        if self.neofs_env.get_binary_version(self.neofs_env.neofs_node_path) > "0.44.2":
```
cthulhu-rider, line 565: offtop: how often we cmp versions like this? May break upon reaching `0.100.0`. Still better to check nums

pytest_tests/tests/object/test_object_api.py
```diff
@@ -562,6 +562,28 @@ def test_object_get_range_hash(
...
+            with allure.step("Verify zero payload ranges"):
+                range_hash = get_range_hash(
+                    wallet,
+                    cid,
+                    oid,
```
cthulhu-rider, line 570: will both simple and complex objects be tested here?

evgeniiz321: yes, this test runs on both types

pytest_tests/tests/object/test_object_api.py
evgeniiz321: good point, i removed all versions comparisons right now, since we updated the node to a new version

evgeniiz321: but looks like 0.45.0 is not fresh enough for get_range to work. Replaced this comparison with a usage of `from packaging.version import Version` it is a more robust standard python way to compare versions.

evgeniiz321: replaced with a new parse_version function to properly compare versions

pytest_tests/tests/object/test_object_api.py
```diff
@@ -562,27 +562,26 @@ def test_object_get_range_hash(
                         f"Expected range hash to match {range_cut} slice of file payload"
                     )
 
-        if self.neofs_env.get_binary_version(self.neofs_env.neofs_node_path) > "0.44.2":
```
roman-khimov: 0:0 range only works correctly post-0.45.0

pytest_tests/tests/object/test_object_searchv2.py
```diff
@@ -40,7 +40,7 @@ def get_attribute_value_from_found_object(found_object: dict, attr_name: str) ->
 def container(default_wallet: NodeWallet, client_shell: Shell, neofs_env: NeoFSEnv) -> str:
     cid = create_container(default_wallet.path, shell=client_shell, endpoint=neofs_env.sn_rpc, rule="REP 3 CBF 3")
     yield cid
-    # delete_container(default_wallet.path, cid, shell=client_shell, endpoint=neofs_env.sn_rpc)
+    delete_container(default_wallet.path, cid, shell=client_shell, endpoint=neofs_env.sn_rpc)
```
roman-khimov, line 45: Why was it disabled? Why is it enabled?

pytest_tests/tests/object/test_object_api.py
evgeniiz321: get_range_hash test works ok for 0.45.0 as well, so i skipped only regular get_range test

pytest_tests/tests/object/test_object_searchv2.py
evgeniiz321: it was disabled for debug purposes and i forgot to uncomment in one of my previous PRs

pytest_tests/tests/object/test_object_api.py
roman-khimov: ok

pytest_tests/tests/object/test_object_searchv2.py
roman-khimov: This should be written somewhere in the commit message.

evgeniiz321: Updated

roman-khimov (approved):

PR #966, opened 2025-02-28
NeoFS 0.45.0 and other updates

roman-khimov:


cthulhu-rider (approved):

roman-khimov:
Meta resync test fails here, but it doesn't seem to be a test issue.

PR #969, opened 2025-03-01
tests: prevent creation of 2 envs for failover tests

evgeniiz321:
fixture simple_object_size depends on the session
scope neofs_env fixture that deploys neofs env
that is not needed by failover tests since they
use their own function scope neofs env


roman-khimov:
Still failing.

evgeniiz321:
It is not the fix for test_enable_resync_metabase_delete_metadata. It is just a minor improvement for local runs - to prevent starting of 2 envs instead of 1.

roman-khimov (approved):

PR #971, opened 2025-03-03
.github: fetch tags during checkout to generate version

evgeniiz321:


.github/workflows/system-tests.yml
```diff
@@ -179,6 +179,8 @@ jobs:
             with:
               repository: nspcc-dev/neofs-node
               ref: ${{ inputs.neofs_node_commit }}
+              fetch-depth: 0
+              fetch-tags: true
```
roman-khimov, line 183: Isn't it the same for all other repos as well? At least gateways.

roman-khimov (approved):

PR #988, opened 2025-03-18
tests: add test for get request forwarding

### LINKED ISSUE #448: [NEW TEST]: Request forwarding (without data duplication)
## Context

A [bug](https://github.com/nspcc-dev/neofs-node/issues/2047) related to request forwarding was found and [fixed](https://github.com/nspcc-dev/neofs-node/pull/2122). Request forwarding is an optimization: when a node is not able to process `GET` locally (obj not found or any other non-logical error) it just wraps the original request and sends it to another node (that probably could handle that request better) AND does not create a new request from scratch.

## Solved problem

A node has an original stream -- the aim of the operation and a list of the container nodes to which it could try to forward the request. Any object part it receives from a container node is sent to the original requester via the original stream immediately. If any error appears in forwarding, a node repeats the same forwarding process to another node from the beginning and _resends all the object parts it has already sent_. That duplication was solved in the attached PR.

## Test that could fix and check that behavior

1. Create a container that _does not_ include all the nodes in the network (e.g. `... CBF 1 SELECT 2 FROM *` if there are 4 nodes);
2. Put an object huge enough to guarantee that getting that object will take measurable time (to be sure that the test is possible to shutdown a node after `GET` started but before `GET` stopped), it could be big in NeoFS terms;
3. See what nodes store the object; could be done via `neofs-cli object head ... --ttl 1` or via `neofs-cli container nodes` (if `REP` is equal to `CBF` * `SELECT` all the nodes from the output should store the obj); alternatively, you can just make your container static via container policy and know what exactly nodes store the object;
4. Send `GET`/`GETRANGE` request to a node that _does not store_ the object, that node is expected to forward such requests and to stream the responses;
5. Make some nodes that store the object stop, it will break streaming;
6. A node-forwarder should repeat the request to another container node and continue streaming only the parts that _have not been sent yet_, in other words, `GET` and `GETRANGE` should finish correctly, and the res object _must_ contain a correct payload.

evgeniiz321:
closes #448

### DIFF
--- a/pytest_tests/tests/failovers/test_failover_storage.py
+++ b/pytest_tests/tests/failovers/test_failover_storage.py
@@ -1,12 +1,24 @@
 import logging
+import queue
 import random
+import threading
+import time
 
 import allure
 import pytest
-from helpers.complex_object_actions import wait_object_replication
+from helpers.complex_object_actions import (
+    get_nodes_with_object,
+    get_nodes_without_object,
+    wait_object_replication,
+)
 from helpers.container import create_container
 from helpers.file_helper import generate_file, get_file_hash
-from helpers.neofs_verbs import get_object, put_object, put_object_to_random_node
+from helpers.neofs_verbs import (
+    get_netmap_netinfo,
+    get_object,
+    put_object,
+    put_object_to_random_node,
+)
 from helpers.node_management import storage_node_healthcheck, wait_all_storage_nodes_returned
 from helpers.wellknown_acl import PUBLIC_ACL
 from neofs_testlib.env.env import NeoFSEnv, StorageNode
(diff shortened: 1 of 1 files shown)


pytest_tests/tests/failovers/test_failover_storage.py
```diff
@@ -185,3 +197,91 @@ def test_put_get_without_storage_nodes(
...
+        default_wallet,
+        neofs_env_function_scope: NeoFSEnv,
+        after_run_return_all_stopped_storage_nodes,
+    ):
+        self.neofs_env = neofs_env_function_scope
+        self.shell = self.neofs_env.shell
+
+        with allure.step("Create container"):
+            wallet = default_wallet
+            placement_rule = "REP 2"
```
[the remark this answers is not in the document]
evgeniiz321: Updated

pytest_tests/tests/failovers/test_failover_storage.py
```diff
@@ -185,3 +197,91 @@ def test_put_get_without_storage_nodes(
...
+                oid,
+                shell=self.shell,
+                nodes=self.neofs_env.storage_nodes,
+            )
+
+        with allure.step("Send get request to a random node without object and turn off a node with object"):
+            exception_queue = queue.Queue()
+
+            def get_object_from_node():
+                for _ in range(5):
```
carpawell, line 260: can you, please, explain what this `5` (and `2` below) means?

carpawell (approved):
I think that is exactly what was requested. However, it is not the best way to check the fixed bug. But I have no better ideas.

pytest_tests/tests/failovers/test_failover_storage.py
evgeniiz321: It is just to make a test longer to increase the probability of catching something bad. Here we have in parallel 5 get requests and 2 sequential node reboots.

carpawell: yes, but doesn't it mean that if we have 2 nodes without an object and `range(2)` with `random.choice(nodes_with_object)` can make this test sometimes false negative?

evgeniiz321: Don't follow. We have 2 nodes with an object and 2 nodes without an object. We sequentially reboot nodes with an object, so each time at least 1 node will have an object. Where is the problem?

PR #1007, opened 2025-03-28
tests: add estimations and getnep17transfers outputs to payment tests

evgeniiz321:


pytest_tests/tests/payment/test_container_payments.py
```diff
@@ -153,5 +153,5 @@ def test_container_payments(
                     - user_wallet_balance_after_new_epoch
                     - (single_node_gain_per_epoch * replicas_number)
                 )
-                <= 1
+                <= 5
```
roman-khimov, line 156: Is that in datoshis?

evgeniiz321: what are `datoshis`? this is in GAS, e.g.: 
```
COMMAND: ./neofs-cli --config /home/runner/work/neofs-testcases/neofs-testcases/neofs-testcases/test-run-2025-03-28-21-43-32-946936/env_files/neofs-env-2025-03-28-22-23-08/cli_config_sbgmfwsxww.yml accounting balance --wallet '/home/runner/work/neofs-testcases/neofs-testcases/neofs-testcases/test-run-2025-03-28-21-43-32-946936/env_files/neofs-env-2025-03-28-22-23-08/deposit_withdrawal_test_wallet_llcqkipjdi' --rpc-endpoint 'localhost:41501' --address 'NiaSScgVngVniuAootpo3RorL4hXFV8ZrN'
RETCODE: 0

STDOUT:
80.93234205
```
The problem is that sometimes a user wallet balance is lower than we expect. 5 (according to failed tests) is the maximum difference between what we expect and what we actually get.

roman-khimov: 5 GAS is quite a lot, how much do we expect to be transferred for storage payment? Maybe it's better to check for transfers, not exact balance?

evgeniiz321:
@roman-khimov how estimations should work? I get only this:
```
COMMAND: ./neofs-adm --config /home/runner/work/neofs-testcases/neofs-testcases/neofs-testcases/test-run-2025-03-31-21-56-43-469224/env_files/neofs-env-2025-03-31-22-36-09/inner_ring_vahspzkvzg/ir_network_config_ejofmpznud.yml fschain estimations --rpc-endpoint 'http://localhost:53925' --cid 'Dn5JLjrymRe5ZWLreVPceWoFnQK6iKbj4PEHS1af267n' --epoch 13
RETCODE: 0

STDOUT:
No estimations found for Dn5JLjrymRe5ZWLreVPceWoFnQK6iKbj4PEHS1af267n container in 13 epoch
```
during various epochs. Do I need to configure something else somewhere for it to work?

roman-khimov:
Do you query it for the current epoch? It's always one epoch late. @carpawell can tell a bit more specifics.

carpawell:
@evgeniiz321, there is a problem when you push new epoch forcibly: estimations are handled in the middle of an epoch (there should always happen 2 events, pushed by the Alphabet) so if you do it externally, the flow is broken. Can you, please, try to wait for a new epoch naturally for test purposes? If everything workes, we can think of additional estimations forcing API in the `neofs-adm`.

evgeniiz321:
@carpawell we already wait for a new epoch naturally in this test. We force the epoch only in the beginning of the test to ensure no unexpected epoch switch will happen. After that - we just wait for a new epochs to arrive. But the epoch is 20s long, can it be the problem?

pytest_tests/tests/payment/test_container_payments.py
```diff
@@ -122,6 +124,15 @@ def test_container_payments(
             new_epoch = neofs_epoch.wait_until_new_epoch(neofs_env, new_epoch)
 
         with allure.step("Get balances in the beginning of a new epoch"):
+            neofs_env.neofs_adm().fschain.estimations(
+                rpc_endpoint=f"http://{neofs_env.fschain_rpc}", cid=cid, epoch=new_epoch
```
carpawell, line 128: as i understand, you try to check estimations for just a started epoch, right? i think it will always be empty. have you tried the previous one?

carpawell:
> we already wait for a new epoch naturally in this test

oh, I see

> But the epoch is 20s long, can it be the problem?

no, it should not

pytest_tests/tests/payment/test_container_payments.py
evgeniiz321: yes, i tried previous, current, next and all epochs starting from 0

carpawell:
@evgeniiz321, can you try `-2` as an `epoch` arg to neofs-adm (it supports relative epochs)? It seems like you need either more time in epoch N for N-1 report or took N-2 freely
Algo is:
1. in Nth epoch you are free to collect reports for N-1th epoch (it ended, no storage changes, only now we are free to report), and with default config, it will last for 1/4 of the epoch duration (5 secs as I understand in this test)
2. after 1/4 of the epoch, node starts writing results to the contract (needs some time for signature collection, inclusion to some block, etc), it gonna take ~2 sec

So yes, it is reliable only for N-2 epoch, I was not correct. Worked locally for me.

roman-khimov (approved):

PR #1033, opened 2025-04-17
tests: new by_attribute REST behavior

### LINKED ISSUE #1027: Test new by_attribute REST behavior
After https://github.com/nspcc-dev/neofs-rest-gw/pull/279 REST has more deterministic behavior wrt multiple objects with the same attribute (`FilePath` usually). This needs to be tested. See https://github.com/nspcc-dev/neofs-rest-gw/pull/282 as well.

evgeniiz321:
closes #1027

### DIFF
--- a/pytest_tests/lib/helpers/rest_gate.py
+++ b/pytest_tests/lib/helpers/rest_gate.py
@@ -76,6 +76,37 @@ def get_via_rest_gate(
     return resp
 
 
+@allure.step("Head via REST Gate")
+def head_via_rest_gate(
+    cid: str,
+    oid: str,
+    endpoint: str,
+) -> Union[str, requests.Response]:
+    """
+    This function heads given object from REST gate
+    cid:          container id to get object from
+    oid:          object ID
+    endpoint:     REST gate endpoint
+    """
+    request = f"{endpoint}/objects/{cid}/by_id/{oid}"
+
+    verify_options_request(request)
+    resp = requests.head(request, stream=True, timeout=DEFAULT_OBJECT_OPERATION_TIMEOUT)
+
+    if not resp.ok:
+        raise Exception(
+            f"""Failed to head object via REST gate:
+                request: {resp.request.path_url},
+                response: {resp.text},
+                status code: {resp.status_code} {resp.reason}"""
+        )
+
+    logger.info(f"Request: {request}")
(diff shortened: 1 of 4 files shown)


evgeniiz321:
Head request doesn't work as expected - https://github.com/nspcc-dev/neofs-rest-gw/issues/292

pytest_tests/lib/helpers/rest_gate.py
```diff
@@ -284,6 +361,26 @@ def get_object_by_attr_and_verify_hashes(
...
 
 
+@allure.step("Verify object can be head using REST header attribute")
+def head_object_by_attr_and_verify(
+    oid: str,
+    cid: str,
+    attrs: dict,
+    endpoint: str,
+    request_path: Optional[str] = None,
+    request_path_attr: Optional[str] = None,
```
roman-khimov, line 371: These `request_path` parameters are really confusing and they're not really used. There are different functions for different cases (OID/attr) and each of these cases has exactly one proper request path.

pytest_tests/tests/services/rest_gate/test_rest_gate.py
```diff
@@ -389,6 +393,151 @@ def test_put_http_get_http_with_headers(self, attributes: dict, gw_params):
             request_path_attr=f"/get_by_attribute/{cid}/{quote(str(attr_name))}/{attr_value}",
         )
 
+    @pytest.mark.parametrize("default_timestamp", [True, False])
+    def test_get_by_same_attribute(self, gw_params: dict, default_timestamp: bool):
```
roman-khimov, line 397: Something should tell us that it's time-based sorting that we're testing here.

pytest_tests/lib/helpers/rest_gate.py
roman-khimov: We still support old APIs, so old API tests are relevant. But that's for cases where we explicitly tests these APIs, here we care about completely different things.

evgeniiz321: Updated

pytest_tests/tests/services/rest_gate/test_rest_gate.py
evgeniiz321: Updated

PR #1047, opened 2025-04-29
Use unique names

### LINKED ISSUE #1048: Unstable "put object with ACL" test
" Test S3: put object with ACL " fails from time to time:
https://rest.fs.neo.org/7ZVSQHtioVRiDT2btN9xEmQhczXV9vryCEAHUawfD8Sc/640-1745487140/index.html#suites/88281b571957371013039835abd59035/627b52796f48fa84/

smallhive:
Closes #1048

### DIFF
--- a/pytest_tests/tests/s3/test_s3_object.py
+++ b/pytest_tests/tests/s3/test_s3_object.py
@@ -643,8 +643,6 @@ class TestS3Object(TestNeofsS3Base):
         bucket_versioning,
         bucket,
     ):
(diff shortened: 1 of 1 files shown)


roman-khimov:
Why don't we just use different keys for different objects? Much simpler, logically correct and doesn't require any magic sleeps. The purpose of the test is to check for ACL correctness, using the same name isn't required.

smallhive:
Used unique name for each subtest

roman-khimov (approved):

evgeniiz321 (approved):

PR #1069, opened 2025-05-27
tests: remove MaintenanceModeAllowed

### LINKED ISSUE #1067: Make `MaintenanceModeAllowed` optional in 0.46.1+
See https://github.com/nspcc-dev/neofs-node/pull/3360#issuecomment-2900333836, this field is to be removed, so tests should be adapted to ignore its absence. Since tests are not likely to care about the real value of this particular config it can be version-independent.

evgeniiz321:
closes #1067


evgeniiz321:
@roman-khimov it is a little bit cumbersome to make this parameter version dependent. Do we really want it? If no - I would suggest just to remove it.

PR #1095, opened 2025-07-03
tests: fix message assertion for expired bearer

### LINKED ISSUE #1091: Expired bearer token in tests
## Current Behavior
A number of ACL tests failing with
> output: rpc error: finish object stream: status: code = 1024 message = bearer token has expired
 to match code = 2048.*message = access to object operation denied

## Expected Behavior
Green.

## Possible Solution
Fix epochs.

## Steps to Reproduce
https://rest.fs.neo.org/HXSaMJXk2g8C14ht8HSi7BBaiYZ1HeWh2xnWPGQCg4H6/3489-1751369099/index.html#suites/1b3407f7fedf50c0aa57aa62f57cfc79/288ea951532a83eb/

## Regression
It can be related to https://github.com/nspcc-dev/neofs-node/pull/3404, so epochs needs to be checked, but looks like test doesn't control them properly either since it allows token to expire.

evgeniiz321:
closes #1091

### DIFF
--- a/pytest_tests/lib/helpers/object_access.py
+++ b/pytest_tests/lib/helpers/object_access.py
@@ -29,6 +29,7 @@ def can_get_object(
     bearer: Optional[str] = None,
     wallet_config: Optional[str] = None,
     xhdr: Optional[dict] = None,
+    expected_error: str = OBJECT_ACCESS_DENIED,
 ) -> bool:
     with allure.step("Try get object from container"):
         try:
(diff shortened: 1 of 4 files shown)


pytest_tests/lib/helpers/object_access.py
```diff
@@ -76,7 +76,9 @@ def can_put_object(
                 neofs_env=neofs_env,
             )
         except OPERATION_ERROR_TYPE as err:
-            assert error_matches_status(err, OBJECT_ACCESS_DENIED), f"Expected {err} to match {OBJECT_ACCESS_DENIED}"
+            assert error_matches_status(err, OBJECT_ACCESS_DENIED) or error_matches_status(err, EXPIRED_BEARER_TOKEN), (
+                f"Expected {err} to match {OBJECT_ACCESS_DENIED} or {EXPIRED_BEARER_TOKEN}"
```
roman-khimov, line 80: Isn't it a completely different error? What do we expect here in reality? I thought we're checking ACL rules providing (or not) access, not token expiration.

evgeniiz321: Updated. Now we pass an error message from tests.

roman-khimov (approved):

PR #1098, opened 2025-07-08
env deployment fixes

evgeniiz321:


roman-khimov (approved):

roman-khimov:
What's up with tests?

evgeniiz321:
@roman-khimov 
from IR - 
```
2025-07-08T03:44:16.743Z	error	rpcsrv/server.go:3203	Error encountered with rpc request	{"code": -32603, "cause": "websocket users limit reached", "method": "", "params": "[]"}
```
is it something that i can fix?

roman-khimov:
Likely so. You shouldn't have that many websocket connections in your environment. Basically, it's a single one for SN or S3 gateway. The default limit is 64.

evgeniiz321:
Test env is pretty basic - 4 SN + 1 s3 gw + 1 rest gw. There are some tests that restarts SN but that's it. Can it be that IR on macos sometimes looses connection and makes a new one instead of properly reusing the lost connection? Or at least how can I print existing connections?

pytest_tests/lib/helpers/cli_helpers.py
```diff
@@ -21,7 +21,7 @@
 COLOR_OFF = "\033[0m"
 
 
-def _cmd_run(cmd: str, timeout: int = 30) -> str:
+def _cmd_run(cmd: str, timeout: int = 60 * 7) -> str:
```
roman-khimov, line 24: Extreme.

neofs-testlib/neofs_testlib/shell/local_shell.py
```diff
@@ -88,17 +88,34 @@ def _exec_non_interactive(self, command: str, options: CommandOptions) -> Comman
...
+                            start_time,
+                            datetime.now(UTC),
+                            CommandResult(stdout=lsof_output, stderr="", return_code=0),
+                        )
+                    except subprocess.SubprocessError as lsof_exc:
+                        logger.warning(f"Failed to run lsof for debug: {lsof_exc}")
+
             raise RuntimeError(
-                f"Command: {command}\nError:\nreturn code: {exc.returncode}\noutput: {exc.output}"
+                f"Command: {command}\nError:\nReturn code: {exc.returncode}\nOutput: {exc.output}"
```
roman-khimov, line 115: This change is not related to websockets.

neofs-testlib/neofs_testlib_tests/unit/test_ssh_shell.py
```diff
@@ -127,4 +127,4 @@ def test_non_existing_binary(self):
 
         error = format_error_details(exc.exception)
         self.assertIn("Error", error)
-        self.assertIn("return code: 127", error)
+        self.assertIn("Return code: 127", error)
```
roman-khimov, line 130: All of them don't.

pytest_tests/tests/conftest.py
```diff
@@ -63,26 +63,12 @@ def neofs_env(temp_directory, artifacts_directory, request):
...
     neofs_env = get_or_create_neofs_env(
-        request, with_s3_gw=False, with_rest_gw=False, chain_meta_data=params.get("chain_meta_data", False)
+        request, with_s3_gw=True, with_rest_gw=True, chain_meta_data=params.get("chain_meta_data", False)
     )
     yield neofs_env
     neofs_env.finalize(request)
 
 
-@pytest.fixture(scope="session")
-def neofs_env_s3_gw(temp_directory, artifacts_directory, request):
```
roman-khimov, line 73: How does this affect test time? Won't there be a problem of gw problem blocking other tests?

evgeniiz321: ~15 min per each job run + more extensive cpu load (+6 parallel neofs envs per job). This helped to reduce macos timeouts problems, not sure for how much, but started to see timeouts definitely not so often as without this fix. 

Yes, gw problems will block other tests, I'm not sure how often this is the case, but seems to me quite rare. This is a tradeoff which we need to understand if we want to make.

neofs-testlib/neofs_testlib/shell/local_shell.py
evgeniiz321: removed

neofs-testlib/neofs_testlib_tests/unit/test_ssh_shell.py
evgeniiz321: removed

PR #1113, opened 2025-07-24
Node 48

roman-khimov:

PR #1116, opened 2025-07-28
Use dynamic block time

roman-khimov:
Let's test it separately from #1113 now.


roman-khimov:
Good to go to me, 7 IR deployment is fixed in https://github.com/nspcc-dev/neofs-node/pull/3489.

cthulhu-rider (approved):

PR #1121, opened 2025-07-30
tests: add minimal gc test

### LINKED ISSUE #1115: Minimal GC test
I'm always frustrated when we don't know if GC works or not.

Add a test that would delete an object and poll `control object status` afterwards till it's gone.

evgeniiz321:
closes #1115


evgeniiz321:
@roman-khimov how long should we wait for GC to cleanup?

PR #1133, opened 2025-08-12
tests: test PUT of expired object

### LINKED ISSUE #1131: Test PUT of expired object
See https://github.com/nspcc-dev/neofs-node/issues/2185, this should work as https://rest.fs.neo.org/HXSaMJXk2g8C14ht8HSi7BBaiYZ1HeWh2xnWPGQCg4H6/3734-1754894690/index.html#suites/21ade783424f8b7f50534d1bb781c71b/dee2a000aec0507/ is pretty similar, but via the REST gateway.

evgeniiz321:
closes #1131

### DIFF
--- a/pytest_tests/tests/object/test_object_lifetime.py
+++ b/pytest_tests/tests/object/test_object_lifetime.py
@@ -65,3 +65,25 @@ class TestObjectApiLifetime(TestNeofsBase):
         with allure.step("Check object deleted because it expires-on epoch"):
             with pytest.raises(Exception, match=OBJECT_NOT_FOUND):
                 get_object_from_random_node(wallet.path, cid, oid, self.shell, neofs_env=self.neofs_env)
+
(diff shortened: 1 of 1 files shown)


pytest_tests/tests/object/test_object_lifetime.py
```diff
@@ -65,3 +65,25 @@ def test_object_api_lifetime(
...
+        self.ensure_fresh_epoch()
+
+        with pytest.raises(Exception, match=".*object has expired.*"):
+            put_object_to_random_node(
+                wallet.path,
+                file_path,
+                cid,
+                self.shell,
+                neofs_env=self.neofs_env,
+                expire_at=1,
```
carpawell, line 88: do we know exactly what epoch is now? likely not the first one, but still i would prefer getting the actual one and decreasing it for readability

carpawell (approved):

PR #1146, opened 2025-09-03
tests: add estimations verification to payments tests

### LINKED ISSUE #1137: New container estimations by Storage Nodes
I'm always frustrated when we update SN and it is not tested properly since.

We have changed the way container estimations are written to container. This is now more descriptive and less complex: storage nodes put estimations directly to the contract, and the contract always has some average calculated value. We will adopt `neofs-adm` to work with the new estimations in https://github.com/nspcc-dev/neofs-node/pull/3540 and SN's reports will be updated in https://github.com/nspcc-dev/neofs-node/pull/3539.

However, any RPC calls to `neo-go` can be done using regular HTTP requests, so maybe it is even a better way to test neofs without neofs utils. Calls to chains can be placed in a package and reused for other tests.

## Additional context
https://github.com/nspcc-dev/neofs-contract/pull/507, https://github.com/nspcc-dev/neofs-node/issues/3460.

evgeniiz321:
closes #1137


evgeniiz321:
@carpawell can you please then suggest a few tests from your perspective for new revised estimations? to cover changes?

PR #1192, opened 2025-11-17
env: max_time_per_block decreased to 500ms to speedup tests

evgeniiz321:


roman-khimov:
Metrics test needs to be adjusted, blocks are ticking a lot here.

roman-khimov (approved):

PR #1206, opened 2025-11-24
tests: move some single sn tests to a shared env

evgeniiz321:


evgeniiz321:
> Homo hash setting is network-wide. Shouldn't make any big difference, but let's wait a bit for other things to settle.

The purpose is to save time by eliminating the new env deployment (here we replaced neofs_env_single_sn with a shared neofs_env) where possible without a lot of updates in the tests logic.

PR #1210, opened 2025-11-25
tests: add retry for connection errors

evgeniiz321:


roman-khimov (commented):
What is the problem you're trying to solve?

roman-khimov:
We're operating in a quite sterile environment, so request loss is not something I'd expect.

roman-khimov:
My concern is that this can hide some gateway bugs.

PR #1211, opened 2025-11-25
tests: add retry for an object parts deletion

evgeniiz321:


roman-khimov (commented):
Isn't the problem is https://github.com/nspcc-dev/neofs-node/issues/3621 here?

PR #1218, opened 2025-11-27
.github: decrease a number of macos shards

evgeniiz321:


evgeniiz321:
@roman-khimov right now with 2 runners it takes 2 hours to complete tests. Do we really want to run all tests on macos runners? (do you recall any specific issues found only for mac?) Maybe we can extract some subset? And run everything only on ubuntu?

evgeniiz321:
@carpawell both, macos runners are slow in compare with ubuntu's - there were a lot of failures with 'deadline exceeded', so we had to split tests into 3 parts, each runnin on its own runner. And now there are not enough runners - we are waiting too long.

roman-khimov:
There were some Mac-specific problems, but the main beauty of Mac-based tests to me is that they're different. They take about the same 2h now anyway, so I don't see any big regression from using two shards instead of three. Node shortage however is very annoying. So for now I'd opt for using two runners.

.github/workflows/run-tests.yml
```diff
@@ -25,7 +25,7 @@ permissions: write-all
 
 jobs:
   run_system_tests:
-    uses: nspcc-dev/neofs-testcases/.github/workflows/system-tests.yml@master
+    uses: nspcc-dev/neofs-testcases/.github/workflows/system-tests.yml@ezayats/decrease-macos-shards
```
roman-khimov: ?

evgeniiz321: updated, it was to test the flow.

PR #1225, opened 2025-12-04
tests: fix unexpected error from put object via grpc_utils

evgeniiz321:
Sometimes the request landed on a node that doesn't belong to the container, it caused a non related error message.


evgeniiz321:
> What's the error message in this case? It should be mentioned in commit message.

Updated.

PR #1251, opened 2025-12-25
Bump python and neo-mamba versions

evgeniiz321:


pytest_tests/lib/helpers/wallet_helpers.py
```diff
@@ -150,9 +150,7 @@ def create_wallet_with_money(neofs_env_with_mainchain: NeoFSEnv) -> NodeWallet:
 def get_private_key(wallet: NodeWallet) -> bytes:
     neo3_wallet: Wallet = load_wallet(wallet.path, passwords=[wallet.password] * 3)
     acc = neo3_wallet.accounts[0]
-    return acc.private_key_from_nep2(
-        acc.encrypted_key.decode("utf-8"), wallet.password, _scrypt_parameters=acc.scrypt_parameters
-    )
+    return acc.private_key
```
roman-khimov: Also.

evgeniiz321: Updated.

evgeniiz321:
> > make: python3.13: No such file or directory
> 
> Fails to run on ubuntu-latest.

This is expected. Github workflow is used from the master, not from this PR (it passes on macos only because the macos runner has python3.13, but ubuntu doesn't). 
I verified it with the workflow from this branch - it is fine.

PR #1266, opened 2026-01-19
tests: add session token v2 tests

### LINKED ISSUE #823: Cover new session token functionality
Once https://github.com/nspcc-dev/neofs-api/issues/241 is done we'll need to cover all of the new functions.

evgeniiz321:
closes #823

### DIFF
--- a/neofs-testlib/neofs_testlib/cli/neofs_cli/session.py
+++ b/neofs-testlib/neofs_testlib/cli/neofs_cli/session.py
@@ -39,3 +39,53 @@ class NeofsCliSession(CliCommand):
             wallet_password,
             **{param: value for param, value in locals().items() if param not in ["self", "wallet_password"]},
         )
+
+    def create_v2(
+        self,
+        wallet: str,
+        out: str,
+        rpc_endpoint: str,
+        lifetime: Optional[int] = None,
+        expire_at: Optional[int] = None,
+        address: Optional[str] = None,
+        json: Optional[bool] = False,
+        subject: Optional[list[str]] = None,
+        subject_nns: Optional[list[str]] = None,
+        context: Optional[list[str]] = None,
+        origin: Optional[str] = None,
+        final: Optional[bool] = False,
+        force: Optional[bool] = False,
+        shell_timeout: Optional[int] = None,
+    ) -> CommandResult:
+        """
+        Create V2 session token with subjects and multiple contexts.
+
+        V2 tokens support:
+        - Multiple subjects (accounts authorized to use the token)
+        - Multiple contexts (container + object operations)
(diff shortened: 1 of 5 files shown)


neofs-testlib/neofs_testlib/cli/neofs_cli/session.py
```diff
@@ -39,3 +39,50 @@ def create(
...
+        final: Optional[bool] = False,
+        shell_timeout: Optional[int] = None,
+    ) -> CommandResult:
+        """
+        Create V2 session token with subjects and multiple contexts.
+
+        V2 tokens support:
+        - Multiple subjects (accounts authorized to use the token)
+        - Multiple contexts (container + object operations)
+        - No server-side session key storage (no SessionCreate RPC needed)
```
roman-khimov: Not required, but still possible. Ephemeral keys are fun.

pytest_tests/tests/session_token/test_session_token_v2.py
```diff
@@ -0,0 +1,1528 @@
...
+    @pytest.mark.simple
+    def test_v2_session_token_nns_owner_direct(self, default_wallet):
+        """
+        Test V2 session token with NNS subject in owner-direct pattern
+
+        Steps:
+        1. Register NNS domain name for owner wallet
+        2. Add NNS record for the domain
+        3. Create V2 session token with NNS subject for owner
+        4. Verify owner can perform operations using the token
```
End-rey: If I understand correctly, this test does not verify the correctness of the claim. In this case, all requests will go through an ephemeral key, so it does not matter what other subjects are in the token. What can be verified here is that if the node key is placed in nns and then this node is requested (it must be a different node from the one used to create the token), then everything should work. In this case, the object must be created with the node key, not the ephemeral key. You can also try adding several nodes to one nns and check that requests can be made from all nodes.

evgeniiz321: Updated.

pytest_tests/lib/helpers/nns.py
```diff
@@ -0,0 +1,82 @@
...
+            force=True,
+        )
+
+    with allure.step(f"Add TXT record for domain {domain}"):
+        neofs_env.neo_go().contract.invokefunction(
+            contracts_hashes["nns"],
+            rpc_endpoint=f"http://{neofs_env.fschain_rpc}",
+            wallet_config=neo_go_wallet_config,
+            method="addRecord",
+            arguments=f"{domain} 16 string:{wallet.address}",
```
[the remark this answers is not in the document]
evgeniiz321: Updated to use the correct method signature.

PR #1304, opened 2026-03-26
Adjust metrics/quota tests to new metrics

### LINKED ISSUE #1301: Adjust metrics/quota tests to new metrics
See https://rest.fs.neo.org/Aku38wpn2xEqfEPxmwyAa7N3VnAkMV8PEVkpLqwKM1z3/5044-1774359496/index.html and https://github.com/nspcc-dev/neofs-node/pull/3856. The biggest change is that all physical objects are counted now (which includes ts/link/lock) and all payloads are counted as well (link objects do have payload).

evgeniiz321:
closes #1301

### DIFF
--- a/pytest_tests/tests/quota/test_user_quota.py
+++ b/pytest_tests/tests/quota/test_user_quota.py
@@ -637,11 +637,10 @@ class TestUserQuota(TestQuotaBase):
             )
 
         self.tick_epochs_and_wait(1)
-        expected_objects_after_delete = 2 if quota_type == "soft" else 1
         prev_report = self.wait_until_quota_values_reported(
             cid,
             prev_report,
-            expected_objects=expected_objects_after_delete,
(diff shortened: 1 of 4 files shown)


pytest_tests/tests/quota/test_user_quota.py
```diff
@@ -637,11 +637,10 @@ def test_object_delete_and_quota_reclaim(self, unique_wallet: NodeWallet, quota_
             )
 
         self.tick_epochs_and_wait(1)
-        expected_objects_after_delete = 2 if quota_type == "soft" else 1
         prev_report = self.wait_until_quota_values_reported(
             cid,
             prev_report,
-            expected_objects=expected_objects_after_delete,
+            expected_objects=None,
```
carpawell: why is it `None` everywhere now?

PR #1325, opened 2026-04-13
Dump fstree and metabase contents before/after meta resync

evgeniiz321:


roman-khimov:
No longer relevant with https://github.com/nspcc-dev/neofs-node/pull/3969 merged?

PR #1347, opened 2026-04-24
workflows: run Linux-only tests in PRs

roman-khimov:
These tend to be changed often and this lead to MacOS runner starving (all busy with some old version of tests running for a long time). Run MacOS for master only, PRs can be checked with Linux-based ones only.


roman-khimov:
Slightly different variation can be like this: https://github.com/nspcc-dev/neofs-rest-gw/pull/385

evgeniiz321 (approved):

Issue #257, opened 2022-07-08
Check for certain errors not any


EliChin:
Some operations may fail due to a number of reasons. We should check if failure occurs after a **particular expected error**. Now we only check if an operation fails regardless of the reason.
First, we need to recognize return codes. See them in [types.proto](https://github.com/nspcc-dev/neofs-api/blob/master/status/types.proto)
Second, we need to check CLI return codes, which are the following:
- 0 -- OK
- 2 -- ACCESS_DENIED
- 1 others

Error tracking may be discussed and made more convenient for further testing.

roman-khimov:
Needs correct CLI error codes.

PR #1356, opened 2026-04-28
tests: add a test for an incomplete operation status

### LINKED ISSUE #1194: Test for incomplete operation status.
I'm always frustrated when we get things like https://github.com/nspcc-dev/neofs-node/issues/3678 in production.

Add special tests for operations that can return "incomplete" status, that is PUT/DELETE/SEARCH. The easiest way to achieve that would be to have "REP 4" container with four nodes and one of them being down. Depending on timeout configurations there might be a period when some "context cancelled" error would be returned, but eventually we should get Incomplete status.

evgeniiz321:
closes #1194

### DIFF
--- a/pytest_tests/tests/failovers/test_failover_storage.py
+++ b/pytest_tests/tests/failovers/test_failover_storage.py
@@ -1,8 +1,10 @@
 import logging
 import queue
 import random
+import re
 import threading
 import time
+from typing import Any, Callable
 
 import allure
 import pytest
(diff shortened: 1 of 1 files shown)


pytest_tests/tests/failovers/test_failover_storage.py
```diff
@@ -14,19 +16,109 @@
...
 from helpers.wellknown_acl import PUBLIC_ACL
 from neofs_testlib.env.env import NeoFSEnv, StorageNode
+from tenacity import retry, retry_if_exception_type, stop_after_attempt, wait_fixed
 
 logger = logging.getLogger("NeoLogger")
 
 
 class TestFailoverStorage:
+    INCOMPLETE_STATUS_RETRY_DELAY_SEC = 2
+    INCOMPLETE_STATUS_RETRY_ATTEMPTS = 45
```
roman-khimov: Quite a lot. Why?

pytest_tests/tests/failovers/test_failover_storage.py
```diff
@@ -14,19 +16,109 @@
...
+    INCOMPLETE_STATUS_PATTERNS = (
```
roman-khimov, line 37: Refs. #257.

pytest_tests/tests/failovers/test_failover_storage.py
```diff
@@ -288,3 +384,114 @@ def get_object_from_node():
...
+                cid, oid, 4, shell=self.shell, nodes=self.neofs_env.storage_nodes, neofs_env=self.neofs_env
+            )
+
+        with allure.step("Kill one storage node"):
+            dead_node.kill()
+
+        with allure.step("Wait baseline object replicated to alive nodes"):
+            wait_object_replication(cid, oid, 3, shell=self.shell, nodes=alive_nodes, neofs_env=self.neofs_env)
+
+        with allure.step("PUT eventually returns incomplete status"):
```
roman-khimov, line 431: I guess you're trying to avoid the timeout error window ("eventually")?

evgeniiz321: yes, there can be some unexpected messages for some time, but eventually incomplete status arrives

pytest_tests/tests/failovers/test_failover_storage.py
evgeniiz321: just to be safe and avoid flippers, how much are we expected to wait?

roman-khimov: 30s IIRC

roman-khimov: Technically it's configurable, but I'm not sure you want to touch it.

evgeniiz321: decreased

roman-khimov (approved):

PR #1367, opened 2026-05-15
ir: deploy IR nodes with new metadata chain

carpawell:
This is the best i can currently do for https://github.com/nspcc-dev/neofs-node/pull/3984#issuecomment-4450869718. This will also require adjustment when SN part is ready (https://github.com/nspcc-dev/neofs-node/pull/3991) but now it seems that it should be enough for not faulting every test. I would also expect it to work ok for meta chain tests when IR part is merged, since there was no old code removal. @evgeniiz321, check it carefully, please.


evgeniiz321 (approved):

evgeniiz321:
Updated the code to be backward compatible with existing neofs version used in tests.

roman-khimov (approved):

PR #1371, opened 2026-05-22
NeoFS 0.53.0

roman-khimov:


carpawell (approved):

evgeniiz321:
@carpawell `test_replace_ir_node_from_main_chain` fails with IR node refuses to start because of:
```
build meta committee acc: own public key was not found among multisig keys
```
It came with a new node, can it be related to chain_meta_data changes? Need some guidance if the test needs to be adjusted or if it is a bug.

carpawell:
New meta problems, yes. https://github.com/nspcc-dev/neofs-node/pull/4002 should help.

cthulhu-rider (approved):

PR #1374, opened 2026-06-08
github: bump gh push to neofs actions

evgeniiz321:


evgeniiz321:
@roman-khimov reporting in this PR will fail cause it takes actions from the master, not this PR. After the merge - everything should be fine.

roman-khimov (approved):

PR #1377, opened 2026-06-17
tests: test container PUT with embedded EACL

### LINKED ISSUE #1375: Test container PUT with embedded EACL
Test https://github.com/nspcc-dev/neofs-node/pull/4006, it should be possible to set EACL with the same command that creates container. This can also be used by tests that only want some single simple EACL for the whole duration of tests.

evgeniiz321:
closes #1375

### DIFF
--- a/pytest_tests/tests/acl/test_eacl.py
+++ b/pytest_tests/tests/acl/test_eacl.py
@@ -4,6 +4,9 @@ import allure
 import pytest
 from helpers.acl import (
     EACLAccess,
+    EACLFilter,
+    EACLFilters,
+    EACLHeaderType,
     EACLOperation,
     EACLRole,
     EACLRoleExtended,
@@ -16,7 +19,11 @@ from helpers.acl import (
 )
 from helpers.complex_object_actions import wait_object_replication
 from helpers.container import EC_3_1_PLACEMENT_RULE, create_container
-from helpers.container_access import check_full_access_to_container, check_no_access_to_container
+from helpers.container_access import (
+    check_custom_access_to_container,
+    check_full_access_to_container,
+    check_no_access_to_container,
+)
 from helpers.grpc_responses import (
     EACL_CHANGE_PROHIBITED,
     EACL_CHANGE_RPC_ERROR,
@@ -34,6 +41,7 @@ from helpers.object_access import (
     can_put_object,
     can_search_object,
 )
+from helpers.utility import parse_version
 from helpers.wellknown_acl import PUBLIC_ACL, PUBLIC_ACL_F
 from neofs_env.neofs_env_test_base import TestNeofsBase
 from neofs_testlib.env.env import NeoFSEnv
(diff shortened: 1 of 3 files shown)


pytest_tests/tests/acl/test_eacl.py
```diff
@@ -610,3 +618,238 @@ def test_only_owner_can_set_eacl(
                     shell=self.shell,
                     endpoint=self.neofs_env.sn_rpc,
                 )
+
+
+class TestEACLOnContainerCreation(TestNeofsBase):
+    @pytest.fixture(autouse=True)
+    def _skip_old_node(self):
+        if parse_version(self.neofs_env.get_binary_version(self.neofs_env.neofs_node_path)) <= parse_version("0.53.0"):
```
carpawell, line 626: btw, what is the practical case for such checks? PR has been merged, so any actual version (any one that is based on the current master) can be tested already. also, `not supported before neofs-node 0.53.0` is not fully true: this feature is available not with the first commit after 0.53.0 release

pytest_tests/tests/acl/test_eacl.py
```diff
@@ -610,3 +618,238 @@ def test_only_owner_can_set_eacl(
...
+        user_wallet = wallets.get_wallet()
+
+        with allure.step("Build a valid eACL table"):
+            eacl_deny = [EACLRule(access=EACLAccess.DENY, role=EACLRole.OTHERS, operation=op) for op in EACLOperation]
+            eacl_table = create_eacl(
+                cid="", rules_list=eacl_deny, shell=self.shell, wallet_config=user_wallet.config_path
+            )
+
+        with allure.step("Creating a final-bit container with an eACL must be rejected by the Inner Ring"):
+            with pytest.raises(RuntimeError, match="put container rpc error"):
```
carpawell: i think it is possible to deny such requests earlier; there is no need to make an RPC if a container does not support eACL setting. i will open a PR

pytest_tests/tests/acl/test_eacl.py
```diff
@@ -610,3 +618,238 @@ def test_only_owner_can_set_eacl(
...
+        ],
+    )
+    def test_set_eacl_on_container_creation(self, wallets, file_path, placement_rule):
+        user_wallet = wallets.get_wallet()
+        other_wallet = wallets.get_wallet(EACLRole.OTHERS)
+
+        with allure.step("Build eACL table denying all operations for OTHERS"):
+            eacl_deny = [EACLRule(access=EACLAccess.DENY, role=EACLRole.OTHERS, operation=op) for op in EACLOperation]
+            eacl_table = create_eacl(
+                cid="", rules_list=eacl_deny, shell=self.shell, wallet_config=user_wallet.config_path
```
carpawell, line 644: @roman-khimov, what behaviour would you expect if `container create` takes a table with a _predefined_ CID in it that does not correspond to the calculated one for the container?

carpawell (approved):
While technically everything is correct, i think there is no need for full eACL tests; we just need to be sure that after a correct `container create` execution, `container get-eacl` returns the same table. But do not insist.

pytest_tests/tests/acl/test_eacl.py
carpawell: https://github.com/nspcc-dev/neofs-node/pull/4023

pytest_tests/tests/acl/test_eacl.py
roman-khimov: Error of some kind at the CLI level. Can have a flag to override CID with proper one.

pytest_tests/tests/acl/test_eacl.py
roman-khimov: neofs-testcases use stable 0.53.0 node

pytest_tests/tests/acl/test_eacl.py
carpawell: added to https://github.com/nspcc-dev/neofs-node/pull/4023

carpawell: can be tested too

pytest_tests/tests/acl/test_eacl.py
evgeniiz321: Right now these tests mostly run in 2 places - this repo (with fixed 0.53.0 release where tests should pass as well), neofs-node repo - with the latest code available. So, we need to differentiate between these 2 cases, hence the checks.

pytest_tests/tests/acl/test_eacl.py
roman-khimov: Yeah, worth checking.

pytest_tests/tests/acl/test_eacl.py
```diff
@@ -610,3 +618,238 @@ def test_only_owner_can_set_eacl(
...
+                other_wallet.wallet_path,
+                cid,
+                oid,
+                file_path,
+                shell=self.shell,
+                neofs_env=self.neofs_env,
+                deny_operations=[denied_operation],
+            )
+
+    @allure.title("eACL set on creation can target a specific user by {address.value}")
```
roman-khimov: Seems to be excessive, EACL is EACL and we expect that it works the same way as regular one (same structure).

roman-khimov (commented):
Maybe worth checking that this kind of EACL can still be replaced in a usual manner.

pytest_tests/tests/acl/test_eacl.py
evgeniiz321: Updated

pytest_tests/tests/acl/test_eacl.py
evgeniiz321: Added couple new tests

pytest_tests/tests/acl/test_eacl.py
evgeniiz321: Removed

evgeniiz321:
> Maybe worth checking that this kind of EACL can still be replaced in a usual manner.

Added

roman-khimov (approved):

PR #1404, opened 2026-07-18
Allure optimizations

evgeniiz321:


neofs-testlib/neofs_testlib/reporter/command_logging.py
```diff
@@ -26,6 +30,69 @@ def _max_attach_chars() -> int:
...
+
+def report_command(report_fn: Callable[[], None]) -> None:
+    stack = _deferred_stack()
+    if stack:
+        stack[-1]["pending"].append(report_fn)
+    else:
+        report_fn()
+
+
+def retry(*dargs, **dkwargs):
```
roman-khimov: What's the gain from this change? I fear it can hide some useful details.

evgeniiz321: Per the full report the gain is ~10mb. Most of the times retries are just the same command, with the same output. But yes, it is possible that we can lose some details.

PR #1421, opened 2026-08-05
tests: more sighup tests

### LINKED ISSUE #1418: Add more tests for SIGHUP

I'm always frustrated when we have ways to break SIGHUP, but we haven't noticed them.

Add coverage for the following cases:
- A syntactically valid configuration with missing required settings, such as public gRPC or FS chain endpoints, causes a graceful storage node shutdown on SIGHUP without reporting readiness to systemd.
- Runtime reload failures, such as an invalid TLS certificate/key or an occupied gRPC endpoint, cause the same graceful shutdown.
- Invalid node attributes or UN/LOCODE values are handled through the same graceful shutdown path.

## Additional context

https://github.com/nspcc-dev/neofs-node/pull/4113

evgeniiz321:
closes #1418


roman-khimov (approved):

pytest_tests/tests/system/test_sighup.py
```diff
@@ -114,3 +205,192 @@ def test_sighup_disable_metrics(neofs_env_single_sn: NeoFSEnv):
...
+            ),
+            ["validate configuration", "empty/not set endpoint"],
+        ),
+        (
+            "no_fschain_endpoints",
+            lambda cfg, _env: cfg["fschain"].__setitem__("endpoints", []),
+            ["validate configuration", "no FS chain RPC endpoints"],
+        ),
+        (
+            "incomplete_tls_settings",
```
End-rey: `tls.enabled: true` without `tls.certificate` is a valid self-signed TLS configuration now. Could we rename this case to something like `self_signed_tls_change_requires_restart` and move it out of `invalid_required_settings`? The current name obscures what is actually being tested.

pytest_tests/tests/system/test_sighup.py
```diff
@@ -114,3 +205,192 @@ def test_sighup_disable_metrics(neofs_env_single_sn: NeoFSEnv):
             neofs_env.storage_nodes[0].prometheus_address.split(":")[0],
             neofs_env.storage_nodes[0].prometheus_address.split(":")[1],
         ), "prometheus port is not busy, but should be"
+
+
+def test_sighup_invalid_tls_keeps_public_api(neofs_env_single_sn: NeoFSEnv):
```
End-rey: The test name says that the public API is kept, but the assertion expects the storage node to stop. Please rename it to reflect the actual expected behavior.

pytest_tests/tests/system/test_sighup.py
End-rey: It would also be useful to start the SN with TLS enabled and no configured `tls.certificate`, verify that it starts successfully and uses the generated certificate, and only then check that changing this configuration via SIGHUP requires a restart.

evgeniiz321: Updated

pytest_tests/tests/system/test_sighup.py
evgeniiz321: Updated

End-rey (approved):

